PR #11, opened 2023-04-24
Standalone version of PyCam

ubdbra001:
Updates added:
- `requirements.txt` to track dependencies
- removed `.git` from `ifit_ld` dir so that it is under the `PyCam` version control
- Adjusted paths to account for changes to repo
- `ifit` and `pyplis` installed from forked repos
- Added installation instructions to `README.md` and moved contents of old readme to `notes.md` (may be worth moving the notes into issues)
- Added test spectra data usually included in `ifit` to `doas/calibration`
- removed and re-added test data to fix case problem


ubdbra001:
Need to update the ReadMe so it points to the main PyCam repo rather than my fork. Let me know if you're happy with the changes as they stand and then I'll make those tweaks.

pycam/ifit_ld/ifit_mod/synthetic_suite.py
```diff
@@ -0,0 +1,161 @@
...
+
+@author: Matthew Varnam - The University of Manchester
+@email : matthew.varnam(-at-)manchester.ac.uk
+"""
+
+import numpy as np
+from scipy.interpolate import griddata
+
+from pycam.ifit_ld.ifit.make_ils import make_ils
+from pycam.ifit_ld.ifit.spectral_analysis import Analyser
```
twVolc, line 13: This is possibly a bit messy but I think the Analyser we want to import here is actually `ifit`.Analyser, not ifit_ld.Analyser.

I'm getting an error when trying to run light dilution using `ifit.light_dilution.generate_ld_curves()` from `pycam.doas.ifit_worker`.


Screenshot 2023-05-25 16 28 10

The Analyser from ifit_ld doesn't have fit_window in its fit_window() method, so I think we need to use ifit's Analyser, which does.

Not 100% convinced this will be the last error we see associated with this packages but it might be a start...

ubdbra001: Okay, will update and then we can try it

ubdbra001: Updated now, this looks like the only place this needed to be changed, so should be okay now?

twVolc: Great, I've just had a check and generate_ld_curves() runs without this error so I think this is all good now! The function takes ages to run though, and I don't have time to do a full check that it generates reasonable looking curves right now, so will have to get to this later in the week. Once I'm happy with that I think we're ready to merge this pull request.

twVolc: Ok the curves just finished plotting and this looks good to me, so I think we're all good here.

README.md
```diff
@@ -1,104 +1,34 @@
...
-> 
-> sudo apt install exfat-fuse exfat-utils for SSD compatibility on pi
+## Installation instructions
 
-Hardware setup:
-> Connect GPIO 23 (physical pin 16) on masterpi to GPIO 3 (physical pin 5) on slave pi. This allows off/on functionality through wittypi start-up/shutdown scripts
-> If using a 128GB microSD must expand filesystem after copying disk image. sudo raspi-config > advanced options > expand filesystem. All space should then be available to pi
+1. If you have `git` installed then clone the `standalone` branch of the repo and move into the PyCamPermanent directory:
+```
+git clone -b standalone https://github.com/ubdbra001/PyCamPermanent.git
```
twVolc, line 13: Before the merge should you change this to represent master on my repo? Or will we change this later?

README.md
```diff
@@ -1,104 +1,34 @@
...
+    - Once downloaded you need to extract the files from the zipped file and navigate to the newly extracted directory in your terminal
 
-Dev notes:
-1. If I have an issue with sockets not performing properly, i have recently changed SocketServer.close_socket() to include a 
-try/except clause for shutdown. Previously this contained the shutdown with the clauses. I'm not sure if catching this error
-will change functionality anywhere, but the error was preventing tests from passing and I think throwing an OSError saying the
-socket is no longer connected is not a useful functionality
+2. Create a new `python 3.8` conda environment and activate it:
+```
+conda create -n pycam_standalone python=3.8
```
twVolc, line 22: Similar to the above comment, should we change this to just be "pycam"?

ubdbra001: This doesn't really matter, it's just the name of the conda environment. It deosn't have to match the name of the repo/branch/whatever. If you'd prefer it to be "pycam" then I'm happy to change it.

twVolc:
I've noticed the old run_pycam.py is still in the pycam folder - do we want to remove this to make things tidy, or is there a reason you've kept it in i.e. as a backup if things don't work?

ubdbra001:
> I've noticed the old run_pycam.py is still in the pycam folder - do we want to remove this to make things tidy, or is there a reason you've kept it in i.e. as a backup if things don't work?

I think I left it as a backup, but I don't this we need it anymore.

README.md
twVolc: For simplicity maybe it would be better to shift to "pycam"?

twVolc:
I've looked through everything and I can't see anywhere that would cause issues. From the basic processing tests I've done it all seems to work well. So I think we're ready to merge

ubdbra001:
> Sorry, in trying to test the DEM bug I've found another issue with this branch which I think is probably associated with the module installations, and possibly should be fixed before we merge? Even though this functionality may not work anyway because of the DEM issue. I can't get to testing the DEM issue because there's a Basemap module availability issue. I think this might be related to matplotlib in some way? Not certain though
> 
> Screenshot 2023-06-08 15 18 42

No problem, do you know what version of basemap you've had installed previously?

ubdbra001:
> > > Sorry, in trying to test the DEM bug I've found another issue with this branch which I think is probably associated with the module installations, and possibly should be fixed before we merge? Even though this functionality may not work anyway because of the DEM issue. I can't get to testing the DEM issue because there's a Basemap module availability issue. I think this might be related to matplotlib in some way? Not certain though
> > > Screenshot 2023-06-08 15 18 42
> > 
> > 
> > No problem, do you know what version of basemap you've had installed previously?
> 
> I'm on version 1.3.0, my matplotlib is 3.1.3 if it turns out they're dependent.

Okay, I'll add that to the dependencies, and we'll go from there.

Issue #16, opened 2023-04-25
Generate calibration fit statistics


twVolc:
Some statistics for calibration (column density vs optical depth) need to be generated (plotted and saved) so that the quality of the fit can be generated. Poor quality fits might suggest SO2 camera images are affected by clouds or there are other issues with the retrieval. Such statistics could therefore form the basis for rejecting/ignoring poor quality data,so is an important step for fully-automating data analysis.

twVolc:
Current functionality from the above commit needs improving.
> It will only work properly with 1st order polynomial fits.
 
> Plots update iteratively even when not requested to do so. I think this really slows down processing as the fit stats time series gets quite big.

ubdbra001:
> ...I think to close the issue it would be nice to have these statistics plotted in the GUI whilst processing is running - so that real-time evaluation of these parameters is possible. I did start this in pycam.gui.figures_analysis.DOASFOVSearchFrame() and its fig_cal_params attribute and within its update_plot method. But i didn't ever get it working properly. I think with the arrays you've now generated it should be relatively straightforward to start displaying these parameters within the GUI?

From https://github.com/twVolc/PyCamPermanent/pull/44#issuecomment-1623592021

ubdbra001:
Just to confirm, the panel indicated below is where you'd like the fit statistics plotted:
![image](https://github.com/twVolc/PyCamPermanent/assets/18421270/8cba0e82-410d-413a-9a23-026d8488356e)

Should it be the coefficient for the slope and R squared on the two y-axes?
Should I worry about polynomial regression here or is that somehting to think about later?

twVolc:
Sorry yes I think if you can change the fit error to r-squared would be easier to interpret. We've also just discussed that a legend would be very useful! (My bad). 

I think for now we can ignore issues with second order polynomials for this plot, as probably getting meaningful info from it would be difficult - unless you can think of an obvious way to do this? 

Rearranging the format of this part of the GUI, perhaps so the time series is along the entire bottom, so it's easier to read as it gets longer, would be useful too. But this is probably an enhancement that maybe needs it's own non-priority issue to be opened.

ubdbra001:
Okay, great. I think I've got it and will put in apull request this afternoon.
I was thinking of colouring the labels for each of the y-axes so that they correspond to the colour of the associated points. That way the information is communicated, but without taking up additional space on the plot.

Not sure how to acheive this off of the top of my head, but I can have a think for a future fix.

[...]

PR #44, opened 2023-07-06
Save calibration data

ubdbra001:
Added ability to save cailbration data into csv with headers

### DIFF
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -3314,16 +3314,8 @@ class PyplisWorker:
                 f.write('Calibration offset={}\n'.format(self.doas_cal_adjust_offset))
             f.write('ambient_roi={}\n'.format(self.ambient_roi))
             f.write('Light_dil_cam={}\n'.format(self.got_light_dil))
-            f.write('DOAS_FOV_pos={},{}\n'.format(self.doas_fov_x, self.doas_fov_y))
-            f.write('DOAS_FOV_radius={}\n'.format(self.doas_fov_extent))
-            if self.doas_recal:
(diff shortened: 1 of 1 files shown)


pycam/so2_camera_processor.py
```diff
@@ -3316,6 +3323,46 @@ def save_processing_params(self):
...
+            recal_string = 'DOAS_fov_recal [minutes]=False\n'
+
+        return pos_string + rad_string + remove_string + recal_string
+
+
+    def save_calibration(self):
+        path = os.path.join(self.processed_dir, "full_calibration.csv")
+
+        with open(path, "w") as file:
+            file.write(self.generate_DOAS_FOV_info())
```
twVolc, line 3348: At this point could you use something like
```
fov_string = self.generate_DOAS_FOV_info()
file.write('headerlines={}\n'.format(fov_string.count() + 1))  # Adding 1 to account for the header line itself
file.write(fov_string)
```
This way if we have any further info we want to add to `generate_DOAS_FOV_info()` we can easily do it whilst keeping older files compatible with any way of reading in the data that we create.

ubdbra001: Just had a go with it and that works perfectly. Added in commit 830ff7c.

ubdbra001:
Does this also resolve issue #16?

twVolc:
> Does this also resolve issue #16?

It definitely goes a long way to solving it. I think to close the issue it would be nice to have these statistics plotted in the GUI whilst processing is running - so that real-time evaluation of these parameters is possible. I did start this in `pycam.gui.figures_analysis.DOASFOVSearchFrame()` and its `fig_cal_params` and within its `update_plot` method. But i didn't ever get it working properly. I think with the arrays you've now generated it should be relatively straightforward to start displaying these parameters within the GUI?

PR #47, opened 2023-07-07
Fix bgmode switch

ubdbra001:
Changes what the 'OK' button does in the 'Background model' window, so that the BG model selected is saved and window closed after pressing 'OK'.

Checks to see if the model selected is the same as currently used/loaded.
- If yes then it just closes the window
- If no, then it runs the model (updating the settings, and plot in analysis window) and then closes the window

__NB:__ This only checks that the BG model selection is the same, if other settings are changed but the BG model isn't then these changes will be lost.

Also, changes the label of the 'Run' button to 'Preview'

### DIFF
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -2797,17 +2797,9 @@ class PlumeBackground(LoadSaveProcessingSettings):
 
     def save_and_close(self):
         
-        # WHat is the currently set mode?
-        if pyplis_worker.bg_pycam:
-            current_mode = 7
-        else:
-            current_mode = pyplis_worker.plume_bg_A.mode
-
-        # If the selected mode is different than the current mode, then update current mode
-        if self.bg_mode != current_mode:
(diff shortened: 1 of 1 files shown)


pycam/gui/figures_analysis.py
```diff
@@ -2788,6 +2788,24 @@ def close_window(self):
...
+        
+        # WHat is the currently set mode?
+        if pyplis_worker.bg_pycam:
+            current_mode = 7
+        else:
+            current_mode = pyplis_worker.plume_bg_A.mode
+
+        # If the selected mode is different than the current mode, then update current mode
+        if self.bg_mode != current_mode:
+            self.gather_vars()
```
twVolc, line 2801: Just wondering if there's any reason why we can't have `self.gather_vars()` outside of the if statement - so that whether BG mode has changed or not we ensure that clicking OK will then set all current parameters. Does this sound like logical functionality for an OK button?

PR #53, opened 2023-07-14
Add calibration fit statistics

### LINKED ISSUE #16: Generate calibration fit statistics

ubdbra001:
Fixes #16 

Plot looks like this (although the right y-axis has the label "R squared")
image

Two thoughts:
1. Do you want to change the label for the left y-axis?
    - Fit 1st order doesn't mean much to me, but it may be fine for you
2. Do you want to set limits for the right y-axis? 
    - To me the automatically selected limits of this axis makes small changes in R squared look bigger than they are
    - Could set the axes so they have a specific lower bound that allows you to see true dips in fit (e.g. pulling a number from a hat: 0.7)

### DIFF
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -3584,8 +3584,8 @@ class DOASFOVSearchFrame(LoadSaveProcessingSettings):
         self.ax_cal_params_1 = self.fig_cal_params.subplots(1, 1)
         self.ax_cal_params_2 = self.ax_cal_params_1.twinx()
         # self.fig_cal_params.subplots_adjust(left=0.05, right=0.9, top=0.95, bottom=0.05)
-        self.ax_cal_params_1.set_ylabel('Fit 1st order')
-        self.ax_cal_params_2.set_ylabel('Fit error')
+        self.ax_cal_params_1.set_ylabel('Fit 1st order', color = "red")
+        self.ax_cal_params_2.set_ylabel('R squared', color = "blue")
 
         self.frame_cal_params = ttk.Frame(self.frame_plots, relief=tk.RAISED, borderwidth=3)
         self.cal_params_canvas = FigureCanvasTkAgg(self.fig_cal_params, master=self.frame_cal_params)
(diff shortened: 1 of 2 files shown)


twVolc:
Ah yes labelling the colours through the axes looks good!

1. 1st order coeff. might be better? If you have a better idea I'm happy to go with what you think. I'm just trying to keep it open to working with a 2nd order polynomial, even though I think plotting useful information from that might be less straightforward and require a different format anyway.
2. Yes I think this sounds like a good idea. I'm wondering how 0-1 would look, so we cover everything, that would be the simplest solution. But maybe a more elegant one would be to fix it to something like 0.5-1 but if any data point drops below that we than rescale the axis to display it. This is maybe adding unnecessary extra complexity/processing time, so I'd be tempted to just check how it looks with 0-1. With this dataset it might look a bit boring, but the idea of this plot really is to see when the fit gets awful (which it sometimes does) so we know that data aren't reliable.

ubdbra001:
Okay, changes made with the newest commit.
Here's how it looks:
image

It should be relatively easy to set it up with one lower bound for the y limit (e.g. 0.5), check to see if the $R^2$ values dip below that, and update the lower bound accordingly. Let me know if you'd like to do that before we merge the pull request.

twVolc:
This looks good to me! I think keep the axis limits as they are here. Slightly pedantic but I wonder if a smaller marker size might slightly improve the aesthetics?

ubdbra001:
This is with `markersize = 3`:

image

Does it look better?

twVolc:
Yes this looks good to me. Sorry, now that I look at it there's enough space for the y label to not be abbreviated ( for non-native English speakers this might be a bit clearer). So could you change the label to "1st order coefficient" please. I think we're good to merge after this. Looks great!

ubdbra001:
Okay, updated.
Ready for merging when you're happy.

Issue #43, opened 2023-07-05
Fixing FOV - not certain radius is in correct units


twVolc:
When fixing the FOV of the spectrometer (instead of calculating it by correlating with the optical-depth image stack), I'm not certain that the radius works correctly - i.e. is fixing the radius reproducing exactly the same results as if pyplis calculated the radius. I have a feeling it's possible pyplis units for radius are different to the GUI's (e.g. degrees vs pixels) so we might be causing errors if using the fixed radius option.

To test we would need to compare the optical depth time series for the fixed FOV and for the calibrated/correlated FOV (that uses the pyplis FOV search function). They should be identical if all other processing remains the same.

tpering:
During real-time processing, this will need to be fixed, i.e., we don't want to spend computational time searching for a FOV.

twVolc:
Details on how the FOV is initially generated (and therefore how the parameters are stored) can be found in `PyplisWorker.doas_fov_search` `line 1955` in `master` branch. I believe the FOV data are stored in the pyplis `DoasCalibData` object (in particular probably `DoasCalibData.fov` so it should be as simple as editing specific attributes in there. I think my main current concern is I didn't adequately test this and I think the pycam GUI's FOV definitions (specifically the radius units) might not correctly map to the pyplis FOV units - this is really what needs checking. Note I've only allowed circular FOV definitions - pyplis does allow eliptical FOVs but I think it's unlikely to make a huge difference to the results considering all the other uncertainties involved so sticking to the simpler circular FOV is fine.

twVolc:
Sorry, to sumarise this more succinctly - basically what you should be able to do is set the exact same FOV pixel coordinates and radius as found using the FOV search and get exactly the same calibration data (and thus emission rate time series) out as a result.

twVolc:
I'm not sure the code actually even runs when fixing the FOV at the moment. So this might need more than just getting units right

twVolc:
That sounds like a pretty significant different - as in 1.648 times larger? I would guess it might be the radius that is getting incorrectly defined, as the centres coordinate should be really simple (I hope), it's just the x and y positions of the FOV. 

[...]

PR #65, opened 2023-09-07
Ubdbra001/issue18

ubdbra001:
Current approach is to loading the config file to `pyplis_worker`, which saves it in an attribute as a dictionary (`pyplis_worker.config`). This is then read in by the GUI elements when required (currently only once at startup to load the deafult config).
Still needs:
- [x] Switch default config to `yaml` file
- [x] Loading in user defined config
  - [x] Ensure that GUI is updated with new config
- [x] Ensuring user changes in the GUI are reflected in the `pyplis_worker.config` dict
- [x] Writing the dict on start/finish of the run

There are probably other bits I'm forgetting too.

_Edit:_ 
Need to:
- [x] Save img registration in Processing dir
- [x] Move default_cam_geom path into config file
   - [x] Save default_cam_geom file alongside config
- [x] pull & merge current dev branch

### DIFF
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -1917,7 +1917,7 @@ class PlumeBackground(LoadSaveProcessingSettings):
         """Prepares all tk variables"""
         self.main_gui = main_gui
         self.vars = {'bg_mode': int,
-                     'auto_param': int,
+                     'auto_param_bg': int,
                      'vign_corr': int,
                      'polyfit_2d_mask_thresh': int,
                      'ref_check_lower': float,  # Used to check background region for presence of gas which would hinder background model
@@ -1939,7 +1939,7 @@ class PlumeBackground(LoadSaveProcessingSettings):
 
 
         self._bg_mode = tk.IntVar()
-        self._auto_param = tk.IntVar()
+        self._auto_param_bg = tk.IntVar()
         self._vign_corr = tk.IntVar()
         self._polyfit_2d_mask_thresh = tk.IntVar()
         self._ref_check_lower = tk.DoubleVar()
@@ -2062,8 +2062,8 @@ class PlumeBackground(LoadSaveProcessingSettings):
         self.ref_area_frame.grid(row=0, column=1, sticky='nw', padx=5, pady=5)
 
         row = 0
-        self.auto = ttk.Checkbutton(self.ref_area_frame, text='Automatic reference areas', variable=self._auto_param,
-                                    command=self.update_draw)
+        self.auto = ttk.Checkbutton(self.ref_area_frame, text='Automatic reference areas',
+                                    variable=self._auto_param_bg, command=self.update_draw)
         self.auto.grid(row=row, column=0, columnspan=2, sticky='w')
 
         # Radiobuttons for reference area drawing
@@ -2348,7 +2348,7 @@ class PlumeBackground(LoadSaveProcessingSettings):
         1 = ygrad_rect
         2 = xgrad_rect
         """
-        if self.auto_param:
+        if self.auto_param_bg:
             self.scale_rect_rad.configure(state=tk.DISABLED)
             self.ygrad_rect_rad.configure(state=tk.DISABLED)
             self.xgrad_rect_rad.configure(state=tk.DISABLED)
@@ -2452,12 +2452,12 @@ class PlumeBackground(LoadSaveProcessingSettings):
         self._bg_mode.set(value)
 
     @property
-    def auto_param(self):
-        return self._auto_param.get()
+    def auto_param_bg(self):
+        return self._auto_param_bg.get()
 
-    @auto_param.setter
-    def auto_param(self, value):
-        self._auto_param.set(value)
+    @auto_param_bg.setter
+    def auto_param_bg(self, value):
+        self._auto_param_bg.set(value)
 
     @property
     def vign_corr(self):
@@ -2649,7 +2649,7 @@ class PlumeBackground(LoadSaveProcessingSettings):
             pyplis_worker.plume_bg_A.mode = self.bg_mode
             pyplis_worker.plume_bg_B.mode = self.bg_mode
             pyplis_worker.bg_pycam = False
-        pyplis_worker.auto_param_bg = self.auto_param
+        pyplis_worker.auto_param_bg = self.auto_param_bg
         pyplis_worker.polyfit_2d_mask_thresh = self.polyfit_2d_mask_thresh
         pyplis_worker.ref_check_lower = self.ref_check_lower
         pyplis_worker.ref_check_upper = self.ref_check_upper
@@ -2780,7 +2780,7 @@ class PlumeBackground(LoadSaveProcessingSettings):
             self.bg_mode = 7
         else:
             self.bg_mode = pyplis_worker.plume_bg_A.mode
-        self.auto_param = pyplis_worker.auto_param_bg
+        self.auto_param_bg = pyplis_worker.auto_param_bg
         self.polyfit_2d_mask_thresh = pyplis_worker.polyfit_2d_mask_thresh
         self.ref_check_lower = pyplis_worker.ref_check_lower
         self.ref_check_upper = pyplis_worker.ref_check_upper
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -2857,7 +2857,7 @@ class ProcessSettings(LoadSaveProcessingSettings):
                      'use_sensitivity_mask': int,
                      'use_light_dilution': int,
                      'min_cd': float,
-                     'buff_size': int,
+                     'img_buff_size': int,
                      'save_opt_flow': int,       # If True, optical flow is saved to buffer (takes up more space)
                      'time_zone': int
                      }
@@ -3154,11 +3154,11 @@ class ProcessSettings(LoadSaveProcessingSettings):
         self._min_cd.set(value / 10 ** 16)
 
     @property
-    def buff_size(self):
+    def img_buff_size(self):
         return self._buff_size.get()
 
-    @buff_size.setter
-    def buff_size(self, value):
+    @img_buff_size.setter
+    def img_buff_size(self, value):
         self._buff_size.set(value)
 
     @property
@@ -3269,7 +3269,7 @@ class ProcessSettings(LoadSaveProcessingSettings):
             pyplis_worker.load_BG_img(self.bg_A_path, band='A')
             pyplis_worker.load_BG_img(self.bg_B_path, band='B')
         pyplis_worker.min_cd = self.min_cd
-        pyplis_worker.img_buff_size = self.buff_size
+        pyplis_worker.img_buff_size = self.img_buff_size
         pyplis_worker.save_opt_flow = self.save_opt_flow
         pyplis_worker.time_zone - self.time_zone
 
@@ -3293,7 +3293,7 @@ class ProcessSettings(LoadSaveProcessingSettings):
         self.use_sensitivity_mask = int(pyplis_worker.use_sensitivity_mask)
         self.use_light_dilution = int(pyplis_worker.use_light_dilution)
         self.min_cd = pyplis_worker.min_cd
-        self.buff_size = pyplis_worker.img_buff_size
+        self.img_buff_size = pyplis_worker.img_buff_size
         self.save_opt_flow = pyplis_worker.save_opt_flow
         self.time_zone = pyplis_worker.time_zone
 
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -1727,14 +1727,22 @@ class GeomSettings:
 
     def gather_geom(self):
         """Gathers all geometry data and adds it to the dictionary"""
-        for key in self.geom_dict:
-            self.geom_dict[key] = getattr(self, key)
+        for key in self.pyplis_worker.geom_dict:
+            self.pyplis_worker.geom_dict[key] = getattr(self, key)
 
-    def update_geom(self, show_info=True):
+    def spread_geom(self):
+        for key in self.pyplis_worker.geom_dict:
+            setattr(self, key, self.pyplis_worker.geom_dict[key])
+
+    def update_geom(self, show_info=True, toGUI=False):
         """Updates pyplis MeasGeom object with values from this frame"""
         # Update geometry dictionary and pyplis worker camera object
-        self.gather_geom()
-        self.pyplis_worker.update_cam_geom(self.geom_dict)
+        if toGUI:
+            self.spread_geom()
+        else:
+            self.gather_geom()
+    
+        self.pyplis_worker.update_cam_geom(self.pyplis_worker.geom_dict)
 
         # Update measurement setup with location
         self.volcano_search(show_info)
@@ -1830,7 +1838,7 @@ class GeomSettings:
         self.volcano = self.pyplis_worker.volcano
 
         # Update geometry settings
-        self.update_geom(show_info)
+        self.update_geom(show_info, toGUI=True)
 
     def set_default_instrument_setup(self):
         """Sets default instrument setup to load on startup"""
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -3364,7 +3364,7 @@ class DOASFOVSearchFrame(LoadSaveProcessingSettings):
         self.pyplis_worker.config['max_doas_cam_dif'] = self.max_doas_cam_dif
         self.pyplis_worker.config['polyorder_cal'] = self.polyorder_cal
         self.pyplis_worker.config['fix_fov'] = self.fix_fov
-        self.pyplis_worker.apply_config(subset=self.vars) # Not 100% sure if this is needed here or could be moved into the if-statement
+        self.pyplis_worker.apply_config(subset=self.vars.keys()) # Not 100% sure if this is needed here or could be moved into the if-statement
         if self.fix_fov:
             self.pyplis_worker.doas_fov_x = self.centre_pix_x
             self.pyplis_worker.doas_fov_y = self.centre_pix_y
@@ -4194,13 +4194,13 @@ class CrossCorrelationSettings(LoadSaveProcessingSettings):
         :return:
         """
         self.main_gui = main_gui
-        self.vars = {'cross_corr_recal': int
-                     }
+        self.vars = {'cross_corr_recal': int}
         self._cross_corr_recal = tk.IntVar()
 
     def gather_vars(self):
         # Set cross-correlation recalibration
-        self.pyplis_worker.cross_corr_recal = self.cross_corr_recal
+        self.pyplis_worker.config['cross_corr_recal'] = self.cross_corr_recal
+        self.pyplis_worker.apply_config(subset=self.vars.keys())
 
     def generate_frame(self):
         """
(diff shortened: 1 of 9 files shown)


ubdbra001:
Need to:
- [ ] Save img registration in Processing dir
- [ ] Move cam geom default path into config file
- [ ] pull & merge current dev branch

ubdbra001:
@twVolc I think this is pretty much there. I'm just trying to add saving the cam_geom file alongside the config, but I think it's worth you taking a look at it now to see if there's anything I've missed.

ubdbra001:
Looks like the Optical flow settings aren't saving properly yet. Working on it.

pycam/gui/figures_analysis.py
```diff
@@ -3383,15 +3346,16 @@ def initiate_variables(self, main_gui):
...
+        self.pyplis_worker.config['remove_doas_mins'] = self.remove_doas_mins
+        self.pyplis_worker.config['doas_recal'] = self.doas_recal
+        self.pyplis_worker.config['doas_fov_recal_mins'] = self.doas_fov_recal_mins
+        self.pyplis_worker.config['doas_fov_recal'] = self.doas_fov_recal
+        self.pyplis_worker.config['max_doas_cam_dif'] = self.max_doas_cam_dif
+        self.pyplis_worker.config['polyorder_cal'] = self.polyorder_cal
+        self.pyplis_worker.config['fix_fov'] = self.fix_fov
+        self.pyplis_worker.apply_config(subset=self.vars.keys()) # Not 100% sure if this is needed here or could be moved into the if-statement
         if self.fix_fov:
             self.pyplis_worker.doas_fov_x = self.centre_pix_x
```
twVolc, line 3360: Are these DOAS FOV data still saved somewhere - i.e. in the old processing_settings.txt they are saved as `DOAS_FOV_pos` and `DOAS_FOV_radius`. Really they should be added to the config file, so that if `fix_fov` is True these FOV settings are used to fix the FOV automatically. BUT this relates to #43 - I'm not 100% certain fixing the FOV definitely reproduces the exact FOV parameters as if the FOV was found using the FOV search. If you could check this with a test to ensure you get the exact same emission rate output if fixing the FOV with the defined coordinates and `fov_extent` as you get when using the FOV search, that would be really useful. Then these parameters can be propery incorporated into the config file.

pycam/gui/figures_analysis.py
```diff
@@ -3383,15 +3346,16 @@ def initiate_variables(self, main_gui):
...
-        self.pyplis_worker.doas_cal_adjust_offset = self.doas_cal_adjust_offset
-        self.pyplis_worker.maxrad_doas = self.maxrad_doas
-        self.pyplis_worker.remove_doas_mins = self.remove_doas_mins
-        self.pyplis_worker.doas_recal = self.doas_recal
-        self.pyplis_worker.doas_fov_recal_mins = self.doas_fov_recal_mins
-        self.pyplis_worker.doas_fov_recal = self.doas_fov_recal
-        self.pyplis_worker.max_doas_cam_dif = self.max_doas_cam_dif
-        self.pyplis_worker.polyorder_cal = self.polyorder_cal
-        self.pyplis_worker.fix_fov = self.fix_fov
+        self.pyplis_worker.config['doas_cal_adjust_offset'] = self.doas_cal_adjust_offset
```
twVolc, line 3349: Might be totally wrong here, or there could be a reason for explicitly writing it as you have (perhaps `self.vars` sometimes contains not all the variables or too many?) but I'm wondering if all these lines can be acheived more succintly (as well as future-proofing for adding any new variables) by looping through the `vars` dict something like this:
```       
for key in self.vars:
            self.pyplis_worker.config[key] = getattr(self, key)
```
I'm thinking this could apply to all `gather_vars()` methods that do essentially this - although of course some of the `gather_vars()` do extra things that would still need to be incorporated

ubdbra001: I was trying to preserve the original code as much as possible, but this is a good point. I'll take a look at doing this.

twVolc: I don't know why I didn't do this in my original code... it's not crtiical though so don't spend too much time if it's more hassle. It just suddenly came into my head so I thought I'd note it down

pycam/gui/figures_analysis.py
ubdbra001: Technically yes, as process_settings.txt is still saved, but they can be incorporated into the config file if you think that would be helpful.
The only issue here is that the config file is saved at the start of processing and the `DOAS_FOV_pos` and `DOAS_FOV_radius` aren't determined until the processing run has done the calibration. This isn't necessarily a huge issue, but would require a bit of thinking.

pycam/gui/figures_analysis.py
ubdbra001: There are a couple of bits that could be tidied up (refactoring, removing old code, etc.) might be worth creating an issue for it to keep track and then we can address them at a later date?

twVolc:
I'm just ding some tests. Background modelling doesn't seem to be working properly. I've loaded in a sequence, then loaded a CP registration (which also actually doesn't update until you click No Registration then back to CP registration, but this might be a bug in the main code not this specific branch - I need to test). When I change the ambient ROI in the analysis window this doesn't actually change the backgournd modelling - either automatically (which it should and previously has done) but also even in the Background Model window when clicking Run. When the BG mode is 7, this ambient rectangle should be changing the output - eseentially the image is normalise to this region, so if you draw the ROI over the plume the optical depths should decrease massively, and if you draw it in the clear sky optical depths of the plume should be positive. It seems the ambient ROI isn't being correctly updated in the config file.

pycam/gui/figures_analysis.py
twVolc: Added this as an issue (non-urgent!)

ubdbra001:
> I'm just doing some tests. Background modelling doesn't seem to be working properly. I've loaded in a sequence, then loaded a CP registration (which also actually doesn't update until you click No Registration then back to CP registration, but this might be a bug in the main code not this specific branch - I need to test). When I change the ambient ROI in the analysis window this doesn't actually change the background modelling - either automatically (which it should and previously has done) but also even in the Background Model window when clicking Run. When the BG mode is 7, this ambient rectangle should be changing the output - essentially the image is normalised to this region, so if you draw the ROI over the plume the optical depths should decrease massively, and if you draw it in the clear sky optical depths of the plume should be positive. It seems the ambient ROI isn't being correctly updated in the config file.

I think there is something weird going on with the bg model updating part of the GUI in general. It wasn't updating when I loaded in the config, and needed another run to make it work. Judging by the results I currently thin that it's more cosmetic than anything else (i.e. the final results are the same), but it is still annoying nonetheless.
I spent some time looking into it last week but didn't make any progress, but it is right at the top of my to-do list.

twVolc:
Do the results change if you draw the ROI in a totally different area (e.g. over the plume)? That would prove it's the GUI not updating rather than the model. But it is very useful to be able to redraw the ROI in the analysis window and have it automatically update the SO2 image

ubdbra001:
I haven't tried that, but I'll add it to the things to test when I get back to it.

ubdbra001:
Okay, can confirm that chanaging the ROI on the SO2 image plot does update the underlying ROI value (in pyplis) and the subsequest results it produces.
So, looks like it is just a graphical thing. Completely appreciate that it is a useful feature to have though, so will start looking into what the underlying issue is early next week.

pycam/gui/figures_analysis.py
```diff
@@ -636,13 +636,21 @@ def draw_roi(self, eclick, erelease):
                                                         crop_X, crop_Y, edgecolor='black', fill=False, linewidth=1))
 
         # Only update roi_abs if use_roi is true
-        self.amb_roi = [self.roi_start_x, self.roi_start_y, self.roi_end_x, self.roi_end_y]
+        self.ambient_roi = [self.roi_start_x, self.roi_start_y, self.roi_end_x, self.roi_end_y]
         self.gather_vars()
```
ubdbra001, line 641: Found the issue, needed to apply the updated config dict. An easy enough fix, but this is a pattern I've had to impliment a couple of times.
@twVolc Do you think it would be worth me modifying the `gather_vars` method so that you can specify if you want to apply the config?

twVolc: Yes this sounds smart to me as it seems like the vast majority (if not all?) of the times that `gather_vars` is called it is with the aim of updating the `PyplisWorker` with the new settings. If I remember correctly I think that's essentially what `gather_vars` was desgigned for so perhaps the default `apply_config` should even be `True`, unless you can see a clear reason for it to be `False` by default.

ubdbra001: I think it's the majority of cases, but I'm not certain it is all. So, to err on the side of caution I'd keep the default behaviour of `gather_vars` as it is now (i.e. not running `apply_config`) and only set the flag at times where I know it's needed.

ubdbra001: Okay, added for most appropriate cases. There's one place I need to think a little more about whether/how to include the flag, but it shouldn't affect anything as it stands.

twVolc:
Just been testing things. Loading a new config through File > Load > Load config file doesn't seem to update the camera geometry with the new configuration (in `cam_geom.txt`)

twVolc:
> I'm just doing some tests. Background modelling doesn't seem to be working properly. I've loaded in a sequence, then loaded a CP registration (which also actually doesn't update until you click No Registration then back to CP registration, but this might be a bug in the main code not this specific branch - I need to test). When I change the ambient ROI in the analysis window this doesn't actually change the background modelling - either automatically (which it should and previously has done) but also even in the Background Model window when clicking Run. When the BG mode is 7, this ambient rectangle should be changing the output - essentially the image is normalised to this region, so if you draw the ROI over the plume the optical depths should decrease massively, and if you draw it in the clear sky optical depths of the plume should be positive. It seems the ambient ROI isn't being correctly updated in the config file.

Ignore this - just seen it's an issue so can be looked at at a later date. My memory...

twVolc:
> What are your current thoughts on how best to set defaults? Is the plan to have a button that can set all current settings as the default? Or perhaps a way of selecting a specific config file to become the default? In both cases I guess it involves pulling all the other files along with the yml file - i.e. just dragging the yml file you want into `pycam/conf` and renaming it `processing_setting_defaults.yml` wouldn't work unless the other files are also copied over to that folder.
> 
> This is all probably desirable rather than essential prior to the merge, but I think we discussed that it would be nice to have a button that takes all of the current settings and sets them as the default for start-up. Has that been implemented somewhere?

Ignore this, just seen it's already an issue so isn't needed before the merge. My memory...

pycam/gui/figures_analysis.py
twVolc: Just catching up with things I'd forgotten. Think it's fine to leave this for now and think about it after the merge - fixing #43 and then thinking more about this will probably be something to address in the next part of the work when looking more into real-time processing and smoothing that side of things out

ubdbra001:
> Just been testing things. Loading a new config through File > Load > Load config file doesn't seem to update the camera geometry with the new configuration (in `cam_geom.txt`)

Could you provide a bit more info about what you did? I just had a go with it and didn't see a problem.

twVolc:
> Could you provide a bit more info about what you did? I just had a go with it and didn't see a problem.
> If you could attach the config and the `cam_geom.txt` files that would also help.

I updated the camera geometry - to a new volcano (I've attached the file, although you may need to first download the reventador infor through the GUI for it to recognise this volcano? This is done useing the Get Info button in the camera geometry configuration pane)
[cam_geom.txt](https://github.com/twVolc/PyCamPermanent/files/13242432/cam_geom.txt)
I then saved all of the settings as a new config file (which included saving the new cam_geom.txt file, and this all looks good. I think restrated things, so that villarrica was back, and tried loading the config. Looks like all other things changed (haven't tested checked everything yet) but when opening the camera geometry frame again it is stuck on villarrica. 

I may well have done something wrong, but I've tried a few times and it doesn't seem to be changing.

twVolc:
> > Could you provide a bit more info about what you did? I just had a go with it and didn't see a problem.
> > If you could attach the config and the `cam_geom.txt` files that would also help.
> 
> I updated the camera geometry - to a new volcano (I've attached the file, although you may need to first download the reventador infor through the GUI for it to recognise this volcano? This is done useing the Get Info button in the camera geometry configuration pane) [cam_geom.txt](https://github.com/twVolc/PyCamPermanent/files/13242432/cam_geom.txt) I then saved all of the settings as a new config file (which included saving the new cam_geom.txt file, and this all looks good). I then restarted things, so that villarrica was back, and tried loading the config. Looks like all other things changed (haven't checked everything yet) but when opening the camera geometry frame again it is stuck on villarrica.
> 
> I may well have done something wrong, but I've tried a few times and it doesn't seem to be changing.

Ah I've just looked at the `process_config.yml` file - it doesn't change `default_cam_geom` - it's still set as `./pycam/conf/cam_geom/villarrica_26-03-2018.txt` but really this should change to the `cam_geom` file that is saved alongside all the other config files

twVolc:
> > > Could you provide a bit more info about what you did? I just had a go with it and didn't see a problem.
> > > If you could attach the config and the `cam_geom.txt` files that would also help.
> > 
> > 
> > I updated the camera geometry - to a new volcano (I've attached the file, although you may need to first download the reventador infor through the GUI for it to recognise this volcano? This is done useing the Get Info button in the camera geometry configuration pane) [cam_geom.txt](https://github.com/twVolc/PyCamPermanent/files/13242432/cam_geom.txt) I then saved all of the settings as a new config file (which included saving the new cam_geom.txt file, and this all looks good). I then restarted things, so that villarrica was back, and tried loading the config. Looks like all other things changed (haven't checked everything yet) but when opening the camera geometry frame again it is stuck on villarrica.
> > I may well have done something wrong, but I've tried a few times and it doesn't seem to be changing.
> 
> Ah I've just looked at the `process_config.yml` file - it doesn't change `default_cam_geom` - it's still set as `./pycam/conf/cam_geom/villarrica_26-03-2018.txt` but really this should change to the `cam_geom` file that is saved alongside all the other config files

I'm getting the same issue with ILS_path - this should track updates from View > More Windows > DOAS calibration > Load ILS but new paths aren't saved in the config file

ubdbra001:
> Ah I've just looked at the `process_config.yml` file - it doesn't change `default_cam_geom` - it's still set as `./pycam/conf/cam_geom/villarrica_26-03-2018.txt` but really this should change to the `cam_geom` file that is saved alongside all the other config files

Looks like the save config doesn't save the new cam_geom location. I'll have a look into that tomorrow.

twVolc:
> > > > Could you provide a bit more info about what you did? I just had a go with it and didn't see a problem.
> > > > If you could attach the config and the `cam_geom.txt` files that would also help.
> > > 
> > > 
> > > I updated the camera geometry - to a new volcano (I've attached the file, although you may need to first download the reventador infor through the GUI for it to recognise this volcano? This is done useing the Get Info button in the camera geometry configuration pane) [cam_geom.txt](https://github.com/twVolc/PyCamPermanent/files/13242432/cam_geom.txt) I then saved all of the settings as a new config file (which included saving the new cam_geom.txt file, and this all looks good). I then restarted things, so that villarrica was back, and tried loading the config. Looks like all other things changed (haven't checked everything yet) but when opening the camera geometry frame again it is stuck on villarrica.
> > > I may well have done something wrong, but I've tried a few times and it doesn't seem to be changing.
> > 
> > 
> > Ah I've just looked at the `process_config.yml` file - it doesn't change `default_cam_geom` - it's still set as `./pycam/conf/cam_geom/villarrica_26-03-2018.txt` but really this should change to the `cam_geom` file that is saved alongside all the other config files
> 
> I'm getting the same issue with ILS_path - this should track updates from View > More Windows > DOAS calibration > Load ILS but new paths aren't saved in the config file

`spec_dir` and `dark_spec_dir` also don't seem to change in the config file. Keys that aren't changing when saving a config file:

- `default_cam_geom`
- `ILS_path`
- `spec_dir`
- `dark_spec_dir`

I haven't checked everyone, so there may be more. I guess to be thorough we just need to edit each parameter and be certain it changes when being saved.

ubdbra001:
> `spec_dir` and `dark_spec_dir` also don't seem to change in the config file. Keys that aren't changing when saving a config file:
> 
>     * `default_cam_geom`
> 
>     * `ILS_path`
> 
>     * `spec_dir`
> 
>     * `dark_spec_dir`
> 
> 
> I haven't checked everyone, so there may be more. I guess to be thorough we just need to edit each parameter and be certain it changes when being saved.

Aside from `default_cam_geom` the rest are associated with Doas/Ifit, which I suspect is the problem. Will save looking into that until tomorrow.

twVolc:
> Aside from `default_cam_geom` the rest are associated with Doas/Ifit, which I suspect is the problem. Will save looking into that until tomorrow.

Great! It would be nice to have this working for the DOAS/IFit side of things too as I envisage incorporating new variables into the config for IFit settings in the not-too-distant future - such as `start_fit_wave` and `end_fit_wave` (not for this merge don't worry!)

ubdbra001:
> > Aside from `default_cam_geom` the rest are associated with Doas/Ifit, which I suspect is the problem. Will save looking into that until tomorrow.
> 
> Great! It would be nice to have this working for the DOAS/IFit side of things too as I envisage incorporating new variables into the config for IFit settings in the not-too-distant future - such as `start_fit_wave` and `end_fit_wave` (not for this merge don't worry!)

So, I'm thinking there are two ways to approach this:
1. The quick way - Write a function where a list of the doas params are hardcoded, and then when the config file is saved these values are collected from `doas_worker` and the config dict in `pyplis_worker` is updated. The downside to this is that this list will have to be updated everytime a new doas param is added.
2. The more thorough way - Set up an intermediate dict in `doas_worker` (like I've done with `pyplis_worker`) that can then be passed back to `pyplis_worker` when it comes time to save the config. The downside to this is that it will take some time to go through the code and ensure that this works correctly (especially as I have a poorer grasp of the internals of `doas_worker` than I do `pyplis_woker`).

I'm more inlcined to go for 1 here, as it doesn't rule out doing 2 later and this is already a bit of a hefty PR. What do you think?

twVolc:
> So, I'm thinking there are two ways to approach this:
> 
> 1. The quick way - Write a function where a list of the doas params are hardcoded, and then when the config file is saved these values are collected from `doas_worker` and the config dict in `pyplis_worker` is updated. The downside to this is that this list will have to be updated everytime a new doas param is added.
> 2. The more thorough way - Set up an intermediate dict in `doas_worker` (like I've done with `pyplis_worker`) that can then be passed back to `pyplis_worker` when it comes time to save the config. The downside to this is that it will take some time to go through the code and ensure that this works correctly (especially as I have a poorer grasp of the internals of `doas_worker` than I do `pyplis_worker`).
> 
> I'm more inlcined to go for 1 here, as it doesn't rule out doing 2 later and this is already a bit of a hefty PR. What do you think?

Yeh I think 1 sounds fine to me, I don't think we should embark on anything substantial regarding this right now. It would just be good to get this side of things working. Updating that list to add new params later doesn't sound too arduous.

ubdbra001:
> Yeh I think 1 sounds fine to me, I don't think we should embark on anything substantial regarding this right now. It would just be good to get this side of things working. Updating that list to add new params later doesn't sound too arduous.

Great, I'll do that today.
It's mostly the fact that you need to remember to update the list that makes 1. a slightly worse solution. I don't know how you are, but for me it's easy to forget to update it and then spend ages trying to work out what's going wrong.

twVolc:
> > Yeh I think 1 sounds fine to me, I don't think we should embark on anything substantial regarding this right now. It would just be good to get this side of things working. Updating that list to add new params later doesn't sound too arduous.
> 
> Great, I'll do that today. It's mostly the fact that you need to remember to update the list that makes 1. a slightly worse solution. I don't know how you are, but for me it's easy to forget to update it and then spend ages trying to work out what's going wrong.

Yeh I know what you mean. If you just signpost the list as well as possible (e.g. a comment at the top of the config setup or somewhere you find suitable) hopefully if/when anyone comes to make updates they head for the config section and see how to implement it.

ubdbra001:
Okay, those two commits should ensure the last params related to doas/ifit are are update

Issue #15, opened 2023-04-25
Save calibration fit parameters


twVolc:
Calibration parameters (polynomial coefficients) should be saved for each point in time, to will allow post-hoc interrogation of data. This kind of relates to issue #14.

PR #66, opened 2023-09-14
Load cal

twVolc:
This should make it possible to load in the calibration statistics so that calibration of images is more straightforward for subsequent processing runs. It does mean that the camera images must be processed in exactly the same way (same background modelling procedure) otherwise the calibration parameters don't hold. 

Hopefully this is compatible with the new config plans, or at least will be straightforward enough to incorporate into them.

@ubdbra001 I wonder if it's possible for you to run a quick test on this - e.g. load in calibration stats you have previously generated for the Lascar data, then process images exactly as previously done and see if it produces identical emission rates using your written tests? also as discussed, if you check you're happy with the changes in this pull request that would be great.

Calibration stats can be loaded in Processing > Settings > Calibration method [Preloaded], and selecting the csv file in Processing > Settings > Calibration time series file.

It would be nice to plot these data on the calibration fit statistics plot as soon as they are loaded in, so the user can inspect how calibration changes through time.

### DIFF
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -2857,6 +2857,7 @@ class ProcessSettings(LoadSaveProcessingSettings):
                      'dark_img_dir': str,
                      'dark_spec_dir': str,
                      'cell_cal_dir': str,
+                     'cal_series_path': str,
                      'cal_type_int': int,        # 0 = cell, 1 = doas, 2 = cell + doas
                      'use_sensitivity_mask': int,
                      'use_light_dilution': int,
@@ -2873,8 +2874,9 @@ class ProcessSettings(LoadSaveProcessingSettings):
         self._dark_img_dir = tk.StringVar()
         self._dark_spec_dir = tk.StringVar()
         self._cell_cal_dir = tk.StringVar()
+        self._cal_series_path = tk.StringVar()
         self._cal_type = tk.StringVar()
-        self.cal_opts = ['Cell', 'DOAS', 'Cell + DOAS']
+        self.cal_opts = ['Cell', 'DOAS', 'Cell + DOAS', 'Preloaded']
         self._use_sensitivity_mask = tk.IntVar()
         self._use_light_dilution = tk.IntVar()
         self._min_cd = tk.DoubleVar()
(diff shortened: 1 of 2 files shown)


ubdbra001:
I'll have a play bout with this tomorrow and let you know how it goes

ubdbra001 (commented):
Initial run of the `load_cal` branch is fine and produces the same emission results as the `dev` branch

Running it again but loading in the calibration stats runs fine up until the `save_calibration` part of the script (line 3055 of `so2_camera_processor.py`) where it is looking for `tau_df` and `fit_df`. Both of these are empty and merging 2 empty dfs causes the `Incompatible merge dtype, dtype('float64') and dtype('O'), both sides must have numeric dtype` error that has cropped up a couple of times.

Would be worth adding a conditional to the `save_calibration` function that tells it to write the existing `calibration_series` dataframe rather than trying to (unsuccessfully) generate a new one.

Also, looking at the `calibration_series` dataframe, it has dropped the initial 350 or so rows that don't have r2 values, is this intentional?

ubdbra001:
I added the conditional to the `save_calibration` function that fixes the issue (at least applies a sticking plaster to it), and the emission results are different, but only in places.
The most obvious place is that the results from run 2 start at 2022-05-20 16:05:15 whereas all the other ones start at 2022-05-20 16:05:20. Additionally there are 656 rows in run 2 and only 653 rows in the others. One of those would be the additional row at the ver start but there are two ther rows unaccounted for.
The values for the rows that do match look the same though.

twVolc:
I guess `save_calibration` doesn't need to be run when using this setting as it would/should just regenerate the exact same statistics that have just been loaded in? So really I guess we need a statement to avoid `save_calibration` if `cal_type` is `preloaded`.

> Also, looking at the calibration_series dataframe, it has dropped the initial 350 or so rows that don't have r2 values, is this intentional?

Yes I think I purposefully made it drop any rows that don't have stats as they're not useful. I think actually what needs to be done is edit how the stats are saved and go back through all early rows that have no calibration and assign them to the most recent calibration values - i.e. 30 minutes in or whenever a calibration first becomes available. So this is an issue associated with #15 rather than anything else. But in this update it looks for the nearest available calibration point and uses that - so the first ~30 minutes get calibrated by that first calibration point ~30 minutes in anyway, so it's not a huge issue.

ubdbra001:
> I guess `save_calibration` doesn't need to be run when using this setting as it would/should just regenerate the exact same statistics that have just been loaded in? So really I guess we need a statement to avoid `save_calibration` if `cal_type` is `preloaded`.

My instinct would be to save a copy of `full_calibration.csv` as usual, but just make it a duplicate of the copy loaded in, so that for any proccessing run there is always a `full_calibration.csv` associated with it. But that is more of a personal preferance than anything.

> Yes I think I purposefully made it drop any rows that don't have stats as they're not useful. I think actually what needs to be done is edit how the stats are saved and go back through all early rows that have no calibration and assign them to the most recent calibration values - i.e. 30 minutes in or whenever a calibration first becomes available. So this is an issue associated with #15 rather than anything else. But in this update it looks for the nearest available calibration point and uses that - so the first ~30 minutes get calibrated by that first calibration point ~30 minutes in anyway, so it's not a huge issue.

Gotcha, that should be easy to fix.

ubdbra001:
Okay, there are two options to deal with the bug I found:
1. Skip saving the calibration entirely when the calibration method is [Preloaded]
2. Re-save the preloaded calibration used in that processing run

Do you have a preference @twVolc? I'd lean towards 2 for the sake of completeness of the results, but it could be a pain having all these files around which are effectively identical.

twVolc:
I think you're right, for completeness we should go with 2. Really, we want to be able to go into any processed directory and see exactly how it has been created and interrogate (potentially replot) any products. With the shear amount of data and processed directories we could end up with we don't want to have to hunt around in other places to find parts.

ubdbra001:
Okay, that commit should fix the issue using the second option.
Have a look to see if you're happy with it, and if so feel free to merge.

Issue #62, opened 2023-08-15
Allow user-defined fixed LDF for ifit retrievals


twVolc:
It may be useful to just process all spectra with a fixed LDF - especially once we have a large dataset that may show a relatively stable LDF at certain camera sites. Should be quite straightforward to do - think I just need to add this parameter to the ifit params object, then make a place in the GUI to define the LDF.

PR #83, opened 2023-11-30
Ifit shift updates

twVolc:
Added new iFit functionality:

1.  Can manually shift the spectrum prior to processing - useful for some spectrometers where shifts in calibration are quite large and the iFit built in shift doesn't seem to deal with the issue sufficiently
2. Added functionality to manually set the light dilution factor (LDF) under which all spectra will then be processed. Useful if a location displays quite consistent LDFs - processing this way drastically speeds up processing relative to performing a full LDF calculation for each spectrum.

### DIFF
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -5567,7 +5567,8 @@ class LightDilutionSettings(LoadSaveProcessingSettings):
                      'use_light_dilution_spec': int,
                      'grid_max_ppmm': int,
                      'grid_increment_ppmm': int,
-                     'spec_recal_time': int}
+                     'spec_recal_time': int,
+                     'LDF': float}
 
 
         self.ambient_roi = [0, 0, 0, 0]
@@ -5580,6 +5581,7 @@ class LightDilutionSettings(LoadSaveProcessingSettings):
         self._grid_max_ppmm = tk.IntVar()
         self._grid_increment_ppmm = tk.IntVar()
         self._spec_recal_time = tk.IntVar()
+        self._LDF = tk.DoubleVar()
 
         # Load default values
         self.load_defaults()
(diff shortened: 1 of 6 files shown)


twVolc:
Fixes #62

pycam/gui/figures_analysis.py
```diff
@@ -5852,6 +5862,13 @@ def _setup_spec_frame(self):
         recal_spin.grid(row=1, column=1, sticky='ew', padx=2, pady=2)
 
 
+        # LDF widget
+        tk.Label(options_frame, text='LDF:', font=self.gui.main_font).grid(row=2, column=0, sticky='w', padx=2, pady=2)
+        self.LDF_box = ttk.Spinbox(options_frame, from_=0, to=1, increment=0.01, width=4, textvariable=self._LDF,
+                                   command=lambda: self.gather_vars(rerun_doas=True), font=self.gui.main_font)
+        self.LDF_box.grid(row=2, column=1, sticky='ew', padx=2, pady=2)
+
```
ubdbra001, line 5870: It would be worth changing this so there is something like an 'apply' button. Effectively so that the command is executed once when the final value is selected, rather than on each increment.

ubdbra001:
It doesn't look like changes to the LDF value are currently being saved into the config file. I'm assuming they are meant to be?

ubdbra001 (approved):
This looks good to me.
The only small bug I found is that if you include `use_light_dilution_spec` and `LDF` in a config, i.e.:
```yaml
use_light_dilution_spec: 1
LDF: 0.07
```
and load it in, then the disabled spinbox will show the `LDF` value, not 0.0.
But stepping through the code suggests that the correct value (0.0) is set for processing, so it's not actually causing any problems.

twVolc:
Ok great, I've added a line that should deal with this - just resetting LDF to 0 when the GUI is generated (`if use_light_dilution_spec = True`).

Issue #41, opened 2023-06-28
"Include Calibration Offset" is not saved when ticking "Set as defaults" in the Doas-Calibration window


tpering:
as above, when reloading software, this is unticked when previously it was ticked (maybe this is intended?)

twVolc:
Looks like `processing_setting_defaults.txt` needs the flag `doas_cal_adjust_offset=1` added to it. for neatness mine is under the `# DOAS calibration and DOAS-AA calibration` section. My updates to this file are a mess to put into git because I have so many edits from processing different data (this links to #18). @ubdbra001 I wonder if you can add this update to the `processing_setting_defaults.txt`  file and push this update? Your version of that file might be less corrupted than mine. I'm sure there's a way I can do this without losing my current settings, but i'm just weary of messing things up.

twVolc:
I think maybe if the flag doesn't already exist in `processing_setting_defaults.txt` , it won't be written into the file on saving defaults (this might be a bit of a bug in itself...)

ubdbra001:
Noticed this in current work on config files. Should be fixed in PR #65

Issue #26, opened 2023-06-09
remove bg_dark_A and bg_dark_B in processing_setting_defaults.txt 


twVolc:
These file paths aren't used - all dark images are taken from the `dark_img_dir` so `bg_dark_A` and `bg_dark_B` are obsolete

PR #86, opened 2023-12-01
Merging updates for 2023.12 release

### LINKED ISSUE #3: BG_MODE not changing properly
BG_MODE seems to always be written as 5 in the final processing results. I think if the background processing is run once, it at least changes the BG_MODE during processing, but if Apply is clicked without running it also doesn't change BG_MODE.

### LINKED ISSUE #16: Generate calibration fit statistics

### LINKED ISSUE #18: Add functionality for copying default settings file for user
Need to have some functionality that allows users to copy the defult settings to a local place and then load that into PyCam

### LINKED ISSUE #20: Add functionality for Nadeau plume speed algorithm
Nadeau suggests a different plume speed algorithm which can give image-by-image resolution for plume speeds but using cross correlation. It involves cross correlating a single line profile in successive images - the profile is parallel to plume transport direction, rather than perpendicular. The main issue here would be ensuring the line is in the correct location/orientation, as the plume direction can change quite significantly in short time-scales. Some kind of automated process for drawing this line could work - e.g. using the gas source location and the location of the bulk plume relative to this.

### LINKED ISSUE #27: Error bars are plotted over the data points rather than having the scatter point over the bar [not urgent]
This seems to be an issue that has come from the fact we are now using a newer version of `matplotlib` than the one I had used to develop most of the code (i.e. things plotted more nicely previously). Error bars now seem to plot over the scatter points, both in the spectrometer time series and the DOAS-camera calibration scatter plot. As a result the plots don't look as tidy as for the most part we can't see the blue scatter point itself - just the associated error bar. I wonder if there's a quick way of forcing error bars to the back of the plots or bringing the scatter points forwards?

### LINKED ISSUE #43: Fixing FOV - not certain radius is in correct units

### LINKED ISSUE #62: Allow user-defined fixed LDF for ifit retrievals

### LINKED ISSUE #64: Calibration fit statistic displays wrong coefficient
The plot in DOAS FOV alignment that displays "1st order coefficient" and "R-squared" is displaying the wrong polynomial coefficient for "1st order coefficient". For the line `y=mx+c` this should be showing `m` but it's currently showing `c` and makes it looks like the calibration line is changing quite dramatically. `c` should generally be close to the origin, so has very little affect on the calibration.

### LINKED ISSUE #68: Plume velocity settings not accessible when no optical flow is set on start-up
If all of `flow_raw`, `flow_histo` and `flow_hybrid` are set to 0 (i.e. not to be used) in the config file, the Plume velocity settings frame doesn't load properly - presumably because some processing hasn't occurred on start-up that typically would be performed if any of those settings are set to 1.

```
Exception in Tkinter callback
  File "C:\Users\gg1tcw\AppData\Local\anaconda3\envs\pycam\lib\tkinter\__init__.py", line 1892, in __call__
    return self.func(*args)
  File "C:\Users\gg1tcw\Documents\Python Scripts\PyCamPermanent\pycam\gui\figures_analysis.py", line 4835, in generate_frame
    self._build_fig_vel()
  File "C:\Users\gg1tcw\Documents\Python Scripts\PyCamPermanent\pycam\gui\figures_analysis.py", line 5206, in _build_fig_vel
    self.img_vel = self.pyplis_worker.velo_img
AttributeError: 'PyplisWorker' object has no attribute 'velo_img'
```

### LINKED ISSUE #70: Add set current config as default button

### LINKED ISSUE #79: Unreachable else conditions in `corr_light_dilution`
https://github.com/twVolc/PyCamPermanent/blob/d2c5b7ebbe755fc8c63127714d30148ad0502e99/pycam/so2_camera_processor.py#L1448-L1454

Just been mucking about with a static code analyser (lloking for dead code) and it highlighted that the else conditions for the above statement are unreachable.
This is because in the if statement the 'ON' gets interpreted as True and so the first part of it will always execute (even if `band` is set to a different value.

Likely not the biggest issue but I thought I'd document it.

### LINKED ISSUE #87: cal_series_path will attempt to be loaded regardless of Calibration setting

### LINKED ISSUE #104: Running processing immediately after a previous run seems to overwrite the config file
I just tried running processing immediately after processing data before. All I changed was adding one extra PCS line. When running the processing it didn't create a new `Processed_2` directory, it just overwrit the config file in `Processed_1`. It doesn't all actually run properly (which I guess is a separate issue, as it prints that `No DOAS data point within 2s of image time: 19:51:35. Image is not added to DOAS calibration`. 

The old line emission rate data are still there, so I haven't lost anything, but a `line_3` folder for the new PCS line I drew has been created. This folder s empty (as expected as I stopped early and the processing wasn't processing fully). So for some reason we start to populate the pre-existing processing directory when running immediately after.

### LINKED ISSUE #105: Preloaded calibration in config doesn't change correctly
If preloaded is set as the initial config setting (when loading a config file), and the calibration file is loaded, changing to a different calibration (e.g. DOAS) via GUI settings doesn't actually change things - it is still calibrated by the preloaded data, or at least the preloaded calibration is still the one that is saved. I think it might actually be the latter, as I'm able to open the camera-DOAS calibration window and it looks like that has been running, so that bit maybe is working, but for some reason the config file still saves `cal_series_path` and the calibration series is definitely a copy of the preloaded series as I have tested this with a series from a different time - the `full_calibration` contains the data from the incorrect period of time related to the preloaded data.

ubdbra001:
Checklist for merging:
- [x] Added options to load and save yaml config file (Closes #18, #26, #41) 
- [x] Add option to set which config is loaded on startup (Closes #70)
- [x] Add flow nadeau (Closes #20)
- [x] Generate calibration fit statistics (Closes #16)
- [x] Add ability to load pre-computed calibration fit statistics
- [x] Added ability to shift spectra and to manually set an LDF for processing spectra (Closes #62)
- [x] Fixes #27 
- [x] Fixes #68 
- [x] Fixes #79 
- [x] Fixes #64 
- [x] Fixes #87
- [x] Fixes #3
- [x] Fixes #43
- [x] Fixes #104
- [x] Fixes #105
- [x] Remove unused files
- [x] Final checks to ensure result consistency


twVolc (commented):
I think everything should be fine now, I guess we just wait on your final tests to ensure things are consistent? Or are we waiting on the DOAS FOV work too? The merge conflicts are really small and one of them doesn't even seem to conflict - there's no changed lines in master, only dev. I've tried to resolve the conflicts and save them but it seems I can only make resolve when I'm about to merge?

ubdbra001:
I'll run the tests today/tomorrow just to make sure the results being produced are consistent with previous versions (I don't see any reason why they should have changed), and as part of that I can see what's happening with the fixed FOV work.
I'll let you know if I think it'll take longer than the end of this week, and you can decide wether it's worth delaying to fit it in this merge.

I'll take a look at the merge conflicts later today, and drop you a message if I need any advice on resolving them, but it sounds like ti should be reasonably straightforward.

twVolc:
Very small tidying thing prior to this merge - in the config file could `ambient_roi` be moved to be part of the `# Plume background model` section of the file? Perhaps directly below `bg_mode`. It can be used in light dilution too, which is probably why I put it there in the first place, but I think it will be more clear to users if it's in the `# Plume background model` section.

ubdbra001:
> Very small tidying thing prior to this merge - in the config file could `ambient_roi` be moved to be part of the `# Plume background model` section of the file? Perhaps directly below `bg_mode`. It can be used in light dilution too, which is probably why I put it there in the first place, but I think it will be more clear to users if it's in the `# Plume background model` section.

I was going to suggest that you should maybe have a look and reorder the parameters in the config file so that they are grouped in the most appropriate way. I'm happy to do this if you like, but it should be just a matter of moving them around in the `processing_setting_defaults.yml` file, as the config structure is derived from there.

twVolc:
Ok great, I can have a look at this then.

PR #88, opened 2023-12-01
Cal_series_path_fix

ubdbra001:
Small fix to make sure that load_cal_series is only called when the calibration type is set accordingly

### DIFF
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -490,7 +490,10 @@ class PyplisWorker:
     @cal_series_path.setter
     def cal_series_path(self, value):
         self._cal_series_path = value
-        self.load_cal_series(self._cal_series_path)
+        # This checks against the value in the config dict as there is no guarantee
+        # that the attribute will be updated before reaching this setter.
+        if self.config["cal_type_int"] == 3:
+            self.load_cal_series(self._cal_series_path)
 
     @property
     def flow_glob(self):


twVolc:
Worth double checking whether things stay working if the `cal_type_int` is manually changed in the settings frame - will this then load things even if the `cal_series_path` isn't directly changed at that time. Only thinking this as the loading is done directly through setting the path, so we just need to make sure that loading also occurs if the path doesn't change. Presumably it does - if it doesn't this would be a bug not related to your change here but related to my earlier work on this...

ubdbra001:
I think this works as expected, but I'll describe what it does and you can see if you agree:
- If you start off with the calibration set to 'Preloaded' and a path specified it will load the file to `pyplis_worker.calibration_series`, if you then change the calibration to something else (e.g. DOAS), it won't remove the loaded series, but it will effectively be ignored for processing.
- If you start off with calibration set to something other than 'Preloaded' and then change it to "Preloaded" (and include a file path) then when you press "OK" the file will be loaded into `pyplis_worker.calibration_series`. If you go back and change it the calibration to something else, as before the series will remian loaded but not affect processing.
- If instead you press "Apply" and then either "Cancel" or "X" then the config file will updated with the values that were in the GUI when you pressed "Apply", but these changes don't actually get applied to `pyplis_worker` immediately. But when you start proccessing they will be applied.
  - Not sure if this is ideal behaviour, maybe the config should be applied to `pyplis_worker` at the end of `gather_vars` regardless?

I can't think of any other use cases, but if you have something in mind not covered here then let me know and I'll give it a test.

PR #96, opened 2023-12-11
Add option to set config loaded on startup

ubdbra001:
Allows users to define a yml file that will be loaded when PyCam starts.
GUI elements may need a bit of modification so they look okay.


twVolc:
Reverting to the original default doesn't work. I think because the settings of the new default file seem to overwrite the `./pycam/conf/processing_settings_defaults.yml` file so when you revert back to this file the settings have all changed in it so it doesn't revert back to Villarrica data.

twVolc:
The issue might be that I clicked "Save changes" after selecting the new config file, perhaps I wasn't supposed to? Maybe this then overwrites the current settings into the default settings. Maybe we need to create a `processing_settings_defaults_safe.yml` file that doesn't ever get edited and the Revert back button changes the file name to this one? THat way we always have an untouched default we can change to even if we've edited the other default file?

ubdbra001:
Ah yeah, I see what's happened. Okay, I'll have a quick think about how best to do that.

ubdbra001:
Just a thought: Do you think haveing the save as default buttons is useful anymore?

twVolc:
> Just a thought: Do you think having the 'save as default' buttons is useful anymore?

Arguably not... There could be the odd situation where it would be nice to just update one thing, but I'm not sure how that would all interact with the fact we can now change the location of the config file, so they might be redundant and just a bit confusing now? I think it's probably not worth the effort of changing things right now, but I'll add an issue to think about those buttons.

ubdbra001:
Okay, the revert button now overwrites the `processing_setting_defaults.yml` with the contents of a backup, and then changes the path stored in `self.default_conf_location`.

twVolc:
That revert button works. But the Save changes button now throws an error 
```
    return self.func(*args)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\gui\menu.py", line 551, in <lambda>
    update_butt = ttk.Button(self.load_frame, text='Save changes', command=lambda: self.set_defaults(self.frame))
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\gui\misc.py", line 313, in set_defaults
    self.pyplis_worker.save_config(filename, subset=self.vars.keys())
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\so2_camera_processor.py", line 442, in save_config
    with open(full_path, "w") as file:
FileNotFoundError: [Errno 2] No such file or directory: './pycam/conf/processing_setting_defaults.yml\\process_config.yml'
```

I'm kind of wondering whether we should actually remove the Save changes button and all of the rest of that start-up stuff, and just have start-up defined by the config file that is selected?

ubdbra001:
I just spotted that too, and then spotted another underlying issue: The `gather_vars()` function for `LoadFrame` doesn't actually do anything. So, the options in that config startup GUI window don't actually get saved by pressing the 'Save changes' button.
Either fixing it or removing the buttons is going to require a little time to do so I suppose it will depend on which you think is a better use of my time?

ubdbra001:
Okay, all removed.

twVolc:
This looks good to me. Shall I merge?

ubdbra001:
Sure!

PR #103, opened 2023-12-15
Config tidy

twVolc:
Rearranging lines in the config file so they are under correct headings

@ubdbra001 if you can double check I haven't done anything stupid here that would be good. But it should be relatively straightforward and fine.


ubdbra001 (approved):
Looks fine to me!

PR #122, opened 2024-05-10
Change units and column headings for output files

ubdbra001:
Re-wrote the code to save the emission rates to a csv.
Now starts by converting the existing `EmissionRates` object to a dataframe, and then performing all the subsequent operations on that.
Maybe the `emis_cols` dictionary could be in a better place though?

### DIFF
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -2886,7 +2886,7 @@ class PyplisWorker:
 
                 # Determine emission rate if we have a velocity
                 if vel_glob is not None:
-                    phi, phi_err = det_emission_rate(cd_buff['cds'][i],
+                    phi, phi_err = self.det_emission_rate_kgs(cd_buff['cds'][i],
                                                      vel_glob,
                                                      cd_buff['distarr'][i],
                                                      cd_buff['cd_err'][i],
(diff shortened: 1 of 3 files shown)


ubdbra001:
How the output looks with the change:
image

Note, that the values are truncated in this preview, but in the actual file they still have ~15 digits after the decimal point

twVolc:
> How the output looks with the change: image
> 
> Note, that the values are truncated in this preview, but in the actual file they still have ~15 digits after the decimal point

To be fair, we probably don't need to write to so many decimal places. 5 significant figures, or just a generic 3 decimal places, would be more than enough - the uncertainties in the data wiould always be much above this.

ubdbra001:
I think this is ready to go. I didn't split out the various flow calculations for `calculate_emission_rate()` here but I did add an issue (#125) to address it in the future.
I put the new `det_emission-rate_kgs()` function in the `Pyplis` worker class. Not sure if that's the best place for it (it doesn't need access to any of the attributes or methods in the class so could just be a stand-alone function), if you think there's a better place do let me know.
Also, I truncated all the values in the output to 3 decimal places.

pycam/so2_camera_processor.py
```diff
@@ -3340,6 +3340,13 @@ def calculate_emission_rate(self, img, flow=None, nadeau_speed=None, plot=True):
 
         return self.results
 
+    @staticmethod
+    def det_emission_rate_kgs(*args, **kwargs):
+        """Convert emission rate from g/s to kg/s"""
+        phi, phi_err = det_emission_rate(*args, **kwargs)
+
+        return (phi/1000, phi_err)
```
twVolc, line 3348: I think `phi_err` will also need dividing by 1000 otherwise we'll have the error in g/s (not that this value is often anything other than nan - but in case in the future this is made use of we should make sure we are correct here.

PR #126, opened 2024-05-15
Issue 120/nadeau config fix

### LINKED ISSUE #120: flow_nadeau plume speed doesn't update from config
It seems that loading the flow_nadeau plume speed information doesn't correctly update from the config file until you actually open the cross-correlation GUI frame. So if you load a config file then run the processing straight away the processing will be incorrect (i.e. I get strange plume speeds because the nadeau line isn't correctly orientated), but if you load a config file, then open the cross-correlation GUI frame this seems to then load in the nadeau settings and you get totally different flow_nadeau plume speed outputs when running processing again.

ubdbra001:
Fixes #120 

Removed the rotation from the GUI code and adjusted the backend code to calculate the coordinates properly.
Added getter/setter for `nadeau_line_orientation` attribute. Not needed strictly speaking, but I find constraining the angle of the orientation useful.

Looks like the primary issue was the y coordinate calculation, sign needed to be inverted to account for the fact that 0 at the top of the image.
Also, adjusted the `calc_line_orientation()` function to deal with this change.

### DIFF
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -4584,7 +4584,7 @@ class CrossCorrelationSettings(LoadSaveProcessingSettings):
         """
         Orientation of Nadeau line
         """
-        return 180 - self._nadeau_line_orientation.get()    # Remove 180 so 0 is upwards on plot
+        return self._nadeau_line_orientation.get()
 
     @nadeau_line_orientation.setter
     def nadeau_line_orientation(self, value):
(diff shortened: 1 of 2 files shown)


ubdbra001:
Ran a test and got two different sets of results (same config, 1 run open the cross corr window, other run didn't)

Looks like the error may be even simpler: The previous line isn't removed when config file loaded.
Adding that in and testing.

ubdbra001:
Okay, working now

PR #128, opened 2024-05-22
Issue 116/display and save config

### LINKED ISSUE #116: Spectra reverts to default "Villarrica" on processing DOAS
Select spectra etc:
image

Click OK:

image

Something unusual going on?

image

ubdbra001:
Fixes #116 

This adds a frame to display the currently specified `spec_dir` for DOAS
It should also ensure that the updated `spec_dir` (either thought the GUI or a config) is saved to a new config properly.

Not sure if it resolves the previously discussed reset of the Dark Spectra/ Spectra directories, so give it a try and if it doesn't let me know.

### DIFF
--- a/pycam/doas/doas_worker.py
+++ b/pycam/doas/doas_worker.py
@@ -461,6 +461,8 @@ class DOASWorker:
             ss = self.spec_dict['plume'][0].split('_')[self.spec_specs.file_ss_loc].replace(ss_id, '')
             self.dark_spec = self.find_dark_spectrum(self.dark_dir, ss)
 
+
+        self.dir_info.update_dir()
         # Update plots if requested
         if plot:
             self.fig_spec.update_clear()
--- a/pycam/doas/doas_worker.py
+++ b/pycam/doas/doas_worker.py
@@ -158,6 +158,7 @@ class DOASWorker:
         # Figures
         self.fig_spec = None            # pycam.doas.SpectraPlot object
         self.fig_doas = None            # pycam.doas.DOASPlot object
+        self.dir_info = None
 
         # Results object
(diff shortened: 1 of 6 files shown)


twVolc:
I'm wondering if `DirIndicator` should go in `doas_analysis.py` rather than `misc.py`? This would just keep all of those widgets on that side of the pane in the same place, so probably makes it easier to find if needed.

Also, possibly should be implemented separately to this PR, but just wondering whether we need to bear in mind how things might work if they were to be run outside of a GUI. In this case, the call to `self.dir_info.update_dir()` would raise an error as presumably the `dir_info` attribute wouldn't have been assigned a `DirIndicator` object. Just with that potential future in mind I wonder if it's worth putting a check in before calling `update_dir()` - e.g.:
```
if isinstance(self.dir_info, DirIndicator):
    self.dir_info.update_dir()
```
or a more generic one that doesn't requrie aany knowledge of the `DirIndicator` class's existence (whether this is important I don't know):
```
if self.dir_info is not None:
    self.dir_info.update_dir()
```

Do you think it's worth including either here?

twVolc:
Great. Yep let's go with `doas_analysis.py`, I think if I came back to things in a year or so that would be the first place I would look for the code for that part of the GUI. Maybe right at the top of that file as it's a small class (it will get lost down the bottom). Just worth double checking it all actually runs once you've moved its location - I have found a lot of issues with circular imports in the past, where I've ended up having to put classes in different files because otherwise multiple files end up wanting to import each other and it all just breaks...

ubdbra001:
Do you mean `figures_doas.py`? Can't find `doas_analysis.py`

twVolc:
> Do you mean `figures_doas.py`? Can't find `doas_analysis.py`

Sorry yes haha

PR #130, opened 2024-05-23
Reset stray_pix values on `doas_worker.dir_load()`

### LINKED ISSUE #118: Can't get spectra to process on latest master branch - outputs constant values, inf error
As above, DOAS just prints out constant values, exactly the same process tried on July 2023 version of pycam and no issues, produces viable results, and processes fine. Going to try a range of different options to see what happens, but for the moment can't process the spectra on latest version.

Don't know where the issue is as no error printed out...

ubdbra001:
Fixes #118 

See issue for more details.

### DIFF
--- a/pycam/doas/doas_worker.py
+++ b/pycam/doas/doas_worker.py
@@ -168,6 +168,7 @@ class DOASWorker:
         """Some resetting of object, before processing occurs"""
         # Reset results objecct
         self.results = DoasResults([], index=[], fit_errs=[], species_id='SO2')
+        self.reset_stray_pix()
 
     @property
     def start_stray_wave(self):
--- a/pycam/doas/doas_worker.py
+++ b/pycam/doas/doas_worker.py
@@ -447,6 +447,8 @@ class DOASWorker:
         # Update first_spec flag TODO possibly not used in DOASWorker, check
         self.first_spec = True
 
+        self.reset_stray_pix()
+
         # Get list of all files in directory
         self.spec_dict = self.get_spec_list()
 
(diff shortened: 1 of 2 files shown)


twVolc:
this looks good to me, but I'm just thinking that this won't fix every case - if we have the default spectrum loaded, but then move to acquiring data - which starts a watch directory by calling `IFitWorker.start_watching()`, if the spectra that now come in don't match the default spectrum (or whatever spectrum is currently loaded) I think we'll run into the same issue that we won't get things processed properly. I think if you add your new `reset_stray_pix()` method to `IfitWorker.reset_self()` too this will account for this use case. Does that make sense?

ubdbra001:
I see what you mean, I'll change that now.

Out of interest is the DoasWorker defunct for processing this kind of data (in your mind at least)?
If not, then I think it would be useful to add a parent class for both IfitWorker and DoasWorker that implements their common features, just to make code maintenance easier.

twVolc:
> Out of interest is the DoasWorker defunct for processing this kind of data (in your mind at least)?

At the moment it's completely defunct unfortunately - I don't think it would run at all if we tried to use it as i haven't kept up with updating for quite a log period of time. I've seen you're continuing to add to both, which is great as it would at some point be good if I can get this back to working properly as it can be nice to compare DOAS with iFit. 

But yes your note on pulling out commonalities into a parent class definitely makes sense and would make keeping on top of the DOASWorker much easier. The only issue is overall this is probably quite an undertaking, so finding the time to do it isn't straightforward as I can't justify prioritising DOAS stuff.

ubdbra001:
> The only issue is overall this is probably quite an undertaking, so finding the time to do it isn't straightforward as I can't justify prioritising DOAS stuff.

That's fair enough, but now I know it's something that would be useful it's maybe something that I can return to in future. I'll add an issue for it now. Plus, at least if you know that the core features of DoasWorker work in the same way as IfitWorker then it makes updating it a bit less of a headache.

PR #137, opened 2024-06-04
Menu tidying ahead of real-time-processing work

tpering:
renames some toolbars, moves locations, updated version, removed redundant


twVolc:
How do you feel about me moving the `Calibration` tab into the `Processing Settings` tab as a drop-down list? I think this way we keep all of the processing settings inside a single drop-down menu which makes it easier for a user to follow through incrementally to ensure they've set everything up. 

Cross-correlation should also be part of `Processing Settings` as it has the `flow_nadeau` setup. I've opened an issue to integrate this frame into the `Plume velocity settings` frame, but until that has been done I think we should keep that one in Processing Settings - with a changed name from `Cross-correlation plot` to `Cross-correlation` to indicate that it has settings within it.

I have the code for this written on this branch, so if you're happy with these changes I can push them.


toolbar

tpering:
Yep - happy with those changes, nice and logical.

PR #139, opened 2024-06-04
Update to truncate_path to fix bug

### LINKED ISSUE #138: truncate_path error
`truncate_path` has introduced an error when loading the Settings window, since `cal_series_path` is `None` by default. Perhaps truncate path needs to accept `None` and just return that if so?
```
Exception in Tkinter callback
  File "C:\Users\thom_\anaconda3\envs\pycam\lib\tkinter\__init__.py", line 1892, in __call__
    return self.func(*args)
  File "C:\Users\thom_\Documents\Work\PyCamPermanent\pycam\gui\figures_analysis.py", line 2980, in generate_frame
    self.cal_series_label = ttk.Label(path_frame, text=self.cal_series_path_short, width=self.path_widg_length,
  File "C:\Users\thom_\Documents\Work\PyCamPermanent\pycam\gui\figures_analysis.py", line 3127, in cal_series_path_short
    return truncate_path(self.cal_series_path, self.path_str_length)
  File "C:\Users\thom_\Documents\Work\PyCamPermanent\pycam\utils.py", line 293, in truncate_path
    raise ValueError("path length should be greater than 0")
ValueError: path length should be greater than 0
```

twVolc:
Fixes #138

@ubdbra001 could you please check this fix, since you wrote `truncate_path()` I want to make sure it still acts how you want it to. But with the previous version it was causing an error with the settings frame (and possibly others) so I've made these edits to make the function a little more forgiving with compatible paths.

### DIFF
--- a/pycam/utils.py
+++ b/pycam/utils.py
@@ -288,9 +288,11 @@ def get_spec_time(filename, date_loc=0, date_fmt="%Y-%m-%dT%H%M%S"):
     return spec_time
 
 def truncate_path(path: str, max_length: int) -> str:
-    """Utility function for truncating path when it exeeds a max_length"""
-    if len(path) <= 0:
-        raise ValueError("path length should be greater than 0")
+    """Utility function for truncating path when it exceeds a max_length"""
+    if len(path) == 0 or path is None:
+        return ''
+    elif len(path) < 0:
+        raise ValueError("Error fetching path length")
 
     if max_length <= 0:
         raise ValueError("max_length should be greater than 0")


ubdbra001 (changes requested):
I don't think the `elif` is needed or even possible to get to. So, if you're happy with it returning '' for an empty string or None then that's okay as it is.

Also, be sure to update the test (in `tests/test_utils.py`) to account for these changes. Happy to have a quick chat about how to do that if you need to.

twVolc:
I've made these updates. But not certain how to actually run the tests, so would you be able to run these to check it works as expected please?

ubdbra001 (approved):

twVolc:
Ah yes, god catch! I guess that's why tests are so useful! Are you happy for me to merge now?

ubdbra001:
Had to make a small change to fix an error.
Just FYI regarding tests: You need to have pytest installed and then you can run the command `pytest ./tests/test_utils.py` in the terminal to run them. There is also integration in most IDEs ([Docs for VsCode](https://code.visualstudio.com/docs/python/testing), [Docs for PyCharm](https://www.jetbrains.com/help/pycharm/pytest.html)).

Issue #113, opened 2024-02-06
Error in `get_cross_corr_emissions_from_buff()` after switching days


ubdbra001:
When moving from one day to the next an error is thrown when `get_cross_corr_emissions_from_buff()` is reached for the first time:

```
Exception has occurred: UnboundLocalError
local variable 'vel_glob_err' referenced before assignment

File "C:\Users\Daniel\Projects\PyCamPermanent\pycam\so2_camera_processor.py", line 2909, in get_cross_corr_emissions_from_buff
    res['flow_glob']._velo_eff_err.append(vel_glob_err)
```

https://github.com/twVolc/PyCamPermanent/blob/952b65ead4365f17a87bcbfc3864700a63098ba2/pycam/so2_camera_processor.py#L2867-L2901

`cross_corr_buff` only has a single data point from the previous day and so `vel_glob` and `vel_glob_err` are not retrieved and assigned (`vel_glob = None` by default, hence why the error is not thrown a line earlier)

ubdbra001:
May be because:
- `got_doas_fov` is only set to True at the very start of processing and then reset set to `False` by `reset_self()` at end of each day. Which means `get_emission_rate_from_buffer()` is skipped in `update_doas_calibration()`
- `first_image` is not reset at the start of each day, not sure if this has a direct impact on the above error though.

Or could be a combination of the two.

Issue #46, opened 2023-07-07
Error during processing means that no results files are output - selecting smaller time periods to process


tpering:
If an error is thrown during processing then the final results files are not written, even for time periods which have been correctly processed and are displaying in the Analysis window. Perhaps there is a way of writing iteratively to a file or writing every 30 minutes of data to the folder. I guess a little bit related to this, it would be nice to have the functionality to just process a smaller portion of the data by selecting a time period (nice to have and not a priority).

tpering:
Just to add here, the output of files when selecting a large time period seems to occur when all data is procesed.

PR #143, opened 2024-06-05
Issue 123/change data save frequency

### LINKED ISSUE #123: Switch output files from hourly to daily
Currently the output data files are split into hourly chunks.
This should be reassessed, maybe files should represent days.

If changing this then it would also likely be beneficial to increase how often the data is written to a file (so if there are issues the data already processed is not lost). Either write as the data is produced (but this may be too much I/O) or write hourly chunks to a single file per day?

ubdbra001:
Closes #123 

Data is now saved incrementally.
For Emissions and DOAS calibration this is first saved after calibration has been run and the emissions data has been updated, and then saved for each individual time point after that.

For DOAS data this will be saved for each individual point as it's processed.

### DIFF
--- a/pycam/gui/menu.py
+++ b/pycam/gui/menu.py
@@ -758,6 +758,12 @@ class LoadFrame(LoadSaveProcessingSettings):
         self.pyplis_worker.load_sequence(pyplis_worker.img_dir, plot_bg=False)
         self.doas_worker.load_dir(self.pyplis_worker.spec_dir, prompt=False, plot=True)
 
+        # This will work but doesn't cover all edge cases
+        if self.pyplis_worker.config.get("FTP_output_dir") is not None:
+            ftp_output_dir = getattr(self.pyplis_worker, "FTP_output_dir")
+            cfg.current_dir_img.root_dir = ftp_output_dir
+            cfg.current_dir_spec.root_dir = ftp_output_dir
+
     def reset_pcs_lines(self):
         """Reset current PCS lines"""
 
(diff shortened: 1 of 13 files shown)


ubdbra001:
Also closes #164, #46, and #113 (I think)

ubdbra001:
Testing these changes:

- Single day of data:
  - [ ] Doas results equivalent
  - [ ] Emission results the same

- Two days of data:
  - [ ] Doas results day 1 equivalent
  - [ ] Doas results day 2 equivalent
  - [ ] Emission rate results day 1 equivalent
  - [ ] Emission rate results day 2 equivalent

- Four days of data:
  - [ ] Runs without throwing errors

pycam/gui/menu.py
```diff
@@ -758,6 +758,12 @@ def reload_config(self):
         self.pyplis_worker.load_sequence(pyplis_worker.img_dir, plot_bg=False)
         self.doas_worker.load_dir(self.pyplis_worker.spec_dir, prompt=False, plot=True)
 
+        # This will work but doesn't cover all edge cases
+        if self.pyplis_worker.config.get("FTP_output_dir") is not None:
+            ftp_output_dir = getattr(self.pyplis_worker, "FTP_output_dir")
+            cfg.current_dir_img.root_dir = ftp_output_dir
+            cfg.current_dir_spec.root_dir = ftp_output_dir
+
```
ubdbra001, line 766: This will work but will likely save in the wrong place if a config file with `FTP_output_dir` is loaded prior to a a config file without: the value from the initial file will be retained.
Thinking about it, this is likely a broader problem that may be worth addressing.

Issue #136, opened 2024-06-04
Create parent class for `DoasWorker` and `IfitWorker` that holds common features


ubdbra001:
As title.

ubdbra001:
Some common methods:
`reset_stray_pix()`
`stray_corr_spectra()`

twVolc:
Just to add, in reality, the vast majority of methods in these classes should be duplicates - I just haven't managed to keep on top of maintaining DOASWorker to a useable state. In an ideal world really the only method that would differ across the two is `process_doas()` and then also some related helper methods that might do things inside `process_doas()`. Also the light dilution correction aspects are reliant on iFit, so these won't be present in DOASWorker. Beyond these two big areas there should be a huge amount of common ground in the classes.

PR #150, opened 2024-06-06
Create a `SpecWorker` class to act as parent for `IFitWorker` and `DOASWorker`

ubdbra001:
These changes move the common features of `IFitWorker` and `DOASWorker` into `SpecWorker`, as outlined in #136.

It should also mean that new features added to `SpecWorker` will be available to both child spectrum processing classes, and make code maintenance easier.

There were some methods that were similar between the two, but not exactly the same, and these were left separate. E.g.:
- `start_watching()`
- `start_processing_threadless()`
- `make_doas_results()`
- `rem_doas_results()`
- `find_dark_spectrum()`


twVolc:
Not sure if these notes are useful or if you're purposefully not wanting to include other methods in `SpecWorker`. Bute here's just a summary:

`start_watching()` can be exactly duplicated - following the `IFitWorker` current design. for the print statements, perhaps both `DOASWorker` and `IFitWorker` can have their own attribute `processing_type` which is a string `DOAS` and `iFit` respectively, and this attribute can be inserted into the print strings where relevant, to indicate which worker is printing that statement. Alternatively the statements can be moved back to the `DOASWorker` style, or they could each just have the base `SpecWorker` as the string, this would be fine (prints statements might be removed eventually anyway so this isn't something to get caught up on).

`start_processing_threadless()` again this can be duplicated, but might require a few updates - e.g. `_process_loop()` in `DOASWorker` would need to take the `continuous_save` input. Maybe it could just take it for now but not do anything with it - making that all work feels like it would be a lot more work. `STOP_FLAG` would need to be a `DOASWorker` attribute, so would just need to be made a `SpecWorker` attribute.

`make_doas_results()` should probably just duplicate `IFitWorker`, but I think either way the line 'doas_results.ldfs = [np.nan] * len(stds)' should be changed to  'doas_results.ldfs = [np.nan] * len(column_densities)'. And looking at it I think this should probably be done anyway, as this method would raise an unwanted error if `stds` remains as `None` and we try `len(stds). This is probably a bug waiting to happen...

`rem_doas_results()` - if the above changes are made to `make_doas_results()`, the `IFitWorker` version of this will be compatible with `DOASWorker` as the `ldfs` will just be `NaNs` and that's fine. So again this could be moved to `SpecWorker`.

`find_dark_spectrum()`. Go with the `IFitWorker` version - I'll have worked with this one more recently so there will be a reason I've made those minor changes - probably through finding bugs or undesired results with the `DOASWorker` version.

ubdbra001:
For this I'm going for the really quick wins, and so just leaving these methods alone for now.
But this information is really useful, so I'll add it to another issue and if/when either of us return to it we can use it.

twVolc:
Ok great. Should we be creating a new dev branch or is it fine to go back to the same branch even though that has been merged with master now?

ubdbra001:
I'm happy with it just being added to the same dev branch, I don't think there's any harm in reusing it.

twVolc:
Cool. Are there any iFit worker tests setup at the moment to ensure the processing stays the same? If so and it has passed them I'm happy to merge this

ubdbra001:
No formal tests yet, but I did a run with the data and settings I've been using and it produces identical results to the current release. Might still be worth you giving that a go too, just to cover off any parts I may not be using with my run.

PR #181, opened 2024-07-24
SO2 image save fix

### LINKED ISSUE #180: Save SO2 PNG setting in GUI doesn't work
In File > Save > Options, "Save SO2 PNG" doesn't make the data actually save SO2 PNGs. It seems that the config attribute "save_img_so2" does still link to saving images, as if you manually change this in the config file it will then save the images. But changing the GUI option doesn't do anything.

When manually ticked in the GUI, the `PyplisWorker.save_dict` is still:
`{'img_aa': {'save': False, 'ext': '.npy'}, 'img_cal': {'save': False, 'ext': '.npy'}, 'img_SO2': {'save': False, 'compression': 0}}`

twVolc:
Fixes #180 by applying config file in gather_vars() of SaveFrame()

### DIFF
--- a/pycam/gui/menu.py
+++ b/pycam/gui/menu.py
@@ -847,6 +847,7 @@ class SaveFrame(LoadSaveProcessingSettings):
 
         self.pyplis_worker.config['save_doas_cal'] = self.save_doas_cal
 
+        self.pyplis_worker.apply_config(subset=self.vars.keys())
         if hasattr(self, 'frame'):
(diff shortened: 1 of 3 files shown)


twVolc:
@ubdbra001 can you please just check you're happy with this bug fix before I merge? Very small and I think it makes sense, but always good to have a second pair of eyes as this relates to your config file stuff.

ubdbra001:
This looks fine to me

PR #187, opened 2024-07-25
Fixes saving of calibrated images

### LINKED ISSUE #158: File > Save > Options > Save calibrated images (ticked) throws error
As above, selecting "Save calibrated images" in the format ".mat" throws this error (didn't check .npy). Code runs if option unselected: 

image

twVolc:
Fixes #158

### DIFF
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -872,6 +872,8 @@ class PyplisWorker:
             i += 1
             self.processed_dir = os.path.join(img_dir, self.proc_name.format(i))
 
+        # Create directory for saved images
+        self.saved_img_dir = os.path.join(self.processed_dir, 'saved_images')
 
     def load_sequence(self, img_dir=None, plot=True, plot_bg=True):
         """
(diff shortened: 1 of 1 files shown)


twVolc:
@ubdbra001 just so that you're aware, I'm anticipating that this will cause a merge conflict with your RTP branch as it has a small update to `PyplisWorker.set_processing_directory()`, which is something that you've updated for RTP to create directories with dates/times of processing. This shouldn't be a big issue though as all we need to do is make sure that your new code also defines `self.saved_img_dir` at the end of that method - as below:

`self.saved_img_dir = os.path.join(self.processed_dir, 'saved_images')`

To optimise code we could also make the directory here (at the moment it's checked twice elsewhere on each iteration loop, and will be made if it doesn't exist, so not the most optimal code but it does the job). I haven't done it like this for the time being as if I make the directory immediately it then messes with this branch's check for if the `Processed_X` directory is completely empty, so we end up just making a lot of unnecessary `Processed_X` folders that aren't populated with any data - they just have an empty `saved_images` folder.

ubdbra001 (approved):
Got it, so when it comes to merging in the RTP branch it should just be a matter of: adding the line you highlighted, and maybe moving the creation of the dir into the make_dir if statement.

Otherwise it all looks good as far as I can tell! 👍

PR #195, opened 2024-07-26
Issue 169/handle img load error

### LINKED ISSUE #169: Error thrown on first processing pair attempted during real-time processing
As below, note this is on poor (quite dark) imagery.

image

ubdbra001:
Should fix #169 using the approach discussed in the issue.

Set-up to attempt to load the image a limited number of times before throwing an error and skipping to the next image.
As with #194 there may be some discussion around the number of attempts and how long to wait between attempts, the numbers I picked were out of thin air.

### DIFF
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -41,6 +41,7 @@ import os
 import cv2
 from skimage import transform as tf
 import warnings
+import traceback
 from ruamel.yaml import YAML
 warnings.simplefilter("ignore", UserWarning)
 warnings.simplefilter("ignore", RuntimeWarning)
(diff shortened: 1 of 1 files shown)


twVolc:
I can't seem to push my edits to your fork (maybe it's supposed to be like this or maybe I've done something wrong) but I think we just need to go all out on this error and catch it as it happens. That being said, this isn't catching it exactly as it happens, as this solution should be easier and SURELY it will still work, by replicating the dark image subtraction. Would you be able to add these edits to your branch if you're happy with it? Essentially within `process_pair()` we just try and artificial image subtraction after `get_img()`. We catch any error and raise `FileNotFoundError` to replicate all other errors associated with this. Surely if we're able to subtract a test image from each file then we'll be able to subtract the dark image???
image

twVolc:
I've made these changes to the local version we're running here. I don't want to speak to soon, but this has been running for 1.5 hours now without hitting the error (before we seemed to hit it within 10-20 minutes). So whilst this solution might be excessive, I think we should just go with it. It would be good to get these changes done and merged with dev today if you're able to @ubdbra001? That would mean we can do a new install of the up-to-date dev tomorrow morning and hopefully have things set up to work before we leave.

Issue #174, opened 2024-07-20
Enhancement - automated production of video for calibrated SO2 image (with a ppm scale)


tpering:
As above, perhaps in the Save > Options section it would be useful to be able to combine saved SO2 png's with a scale (as per the analysis window) into a video on output. 

It would also be nice to be able to set the image colour used for the normal SO2 images saved (as per the analysis window, i.e., the seismic output shows really nice definition).

twVolc:
Haven't done this as a video, but this has been implemented as saving the SO2 image figure. Videos should be created manually by combining the images in some other software (not PyCam). This solution seems suitable for RTP and general functionality of this software, users might be interested in having specific images available for display online for example.

PR #199, opened 2024-07-27
Save matplotlib so2 img

### LINKED ISSUE #198: Config file options for save images don't update GUI
Config files loaded on start-up do seem to correctly update the image save options within File > Save > Options. But if you then load a new config file in once the GUI is open, with different save options, if you go back to File > Save > Options the GUI hasn't updated. The configuration itself has updated though, as running new processing saves the config as expected.

twVolc:
Fixes #198 and pretty much acheives #174 - although saving as images rather than a video. The figure is saved in whatever state the analysis window is currently in, so the user will need to select ppmm in that window for ppmm to be saved. 
Future development should make the save option possible so if ppmm is chosen the analysis window is forced to ppmm once a calibration is avaliable, and then these figures begin to be saved.


twVolc:
Accidentally opened PR for master...

PR #202, opened 2024-07-29
Remove set default buttons

### LINKED ISSUE #99: Set as default buttons may be obsolete?
With the recent work #96, buttons that allow the user to Set as default, may now be obsolete and actually just complicate things. It might be worth considering removing all of these and just settings defaults through changing the config file.

twVolc:
Fixes #99 and should fix #155 too (although havent checked this)


twVolc:
As far as I can see, and have tested, these default buttons and the `set_defaults()` method didn't interfere at all with the new way of saving and loading configuration files. So removing should be totally fine. I just want to check with you that that's the case @ubdbra001 before merging.

twVolc:
Small thought - if we remove these default options, does this mean we don't need the `processing_settings_backup.yml` file as there's no way for the user to edit the `processing_settings_default.yml` file anymore? Removing this file isn't critical, just with the idea of being tidy with the code/repo I guess this file would become obsolete and revert to default settings would just simply switch back to `processing_settings_default.yml`. Happy to keep the backup for added safety though if we think that's a good idea.

ubdbra001:
> Small thought - if we remove these default options, does this mean we don't need the `processing_settings_backup.yml` file as there's no way for the user to edit the `processing_settings_default.yml` file anymore? Removing this file isn't critical, just with the idea of being tidy with the code/repo I guess this file would become obsolete and revert to default settings would just simply switch back to `processing_settings_default.yml`. Happy to keep the backup for added safety though if we think that's a good idea.

I think this is correct. In terms of use cases we have:
- A way to create a new config file
- A way to set the path to a new default config file
- A way to revert back to the original config (effectively resetting the path - though this will likely have to be updated)

Is there anything else a user would need to do that I've missed there?

twVolc:
> I think this is correct. In terms of use cases we have:
> 
> * A way to create a new config file
> * A way to set the path to a new default config file
> * A way to revert back to the original config (effectively resetting the path - though this will likely have to be updated)
> 
> Is there anything else a user would need to do that I've missed there?

I think this sounds like every use-case to me, and seems like it would tidy things up a little.

ubdbra001:
> I guess that would be nice and could speed up changing things. Not a critical thing but I'm guessing it will be a very quick/straightforward implementation.

It should be. If it takes more than 10 mins I'll leave it as it's only saving a couple of clicks.

ubdbra001:
Okay, I'm happy with your changes. Take a look at mine and then feel free to merge if you're happy with everything.

ubdbra001:
I managed to muck up the git history, so fixed it and force pushed it. You may need to re-pull everything in this branch. Sorry!

Issue #203, opened 2024-07-29
doas_results fit_err error


twVolc:
![image](https://github.com/user-attachments/assets/e27c1d5e-61ba-4843-af66-c2ebba768d50)

![image](https://github.com/user-attachments/assets/e8ef8500-bb74-4996-84d5-1c01587fbe83)

This is quite a long traceback, so attached in two consecutive screenshots. The error seems to be associated with trying to extract a data point from `IFitWorker.results.fit_err` beyond the length of that array. Why this is happening isn't immediately clear though. 

This error has crashed all processing - DOAS and camera updates have stopped. Only FTP is still running.

twVolc:
The second part of this `dimension 227 but corresponding boolean dimension is 228` probably provides the answer here. It seems this is an issue associated with the fact that `PyplisWorker` and `IFitWorker` are working in separate threads, so `IFitWorker` is probably in `add_doas_results()` exactly as `PyplisWorker.update_doas_calibration()` is trying to extract `IFitWorker.results.fit_errs`. My guess is the `IFitWorker` thread has exectuted the line `self.results.loc[doas_dict['time']] = cd, but not yet appended to `self.results.fit_errs`, so the lengths of results series don't exactly match with the length of the `fit_err` attribute. I may be barking up the wrong tree but this would be my guess.

If this is the case, I think we can solve this issue using `threading.Lock`. If we add a lock object as an `IFitWorker` attribute, we can acquire that lock for the entirety of `add_doas_results()`. Similarly, `PyplisWorker` can acquire that lock for the entirety of `update_doas_calibration()`. If I understand locks correctly I think this will mean only one of these two code blocks will execute at a time, so there's no danger of half updating `IFitWorker.results`. For completeness we might want to look at any time whatsoever that Pyplis worker might interact with `IFitWorker.results` and ensure that it's done using the lock.

I'll start on this update now. @ubdbra001 let me know if you have any thoughts, or if you think the error is totally different to what I'm thinking.

ubdbra001:
I'd agree it looks like it's trying to access something beyond the range of the array. I don't have too much experience dealing with threads, so I'm happy to defer to you on this one.

twVolc:
Hopefully this has been fixed by #205, but I'll leave this open for a while whilst we're checking this.

PR #205, opened 2024-07-29
Issue 203/add doas lock

twVolc:
Hopefully fixes https://github.com/twVolc/PyCamPermanent/issues/203 by adding a lock which is acquired by `PyplisWorker` and `IFitWorker` when trying to access `IFitWorker.results`.

I currently have this running on the Merapi machine, but the camera has stopped for the day, so we won't find out if #203 is hit again until at least tomorrow. If I've understood the error correctly this should solve it though. Fingers crossed...


ubdbra001 (approved):
This looks sensible to me.
Not sure how to test it beyond just letting it run and seeing if the error pops up again though

PR #206, opened 2024-07-29
Stop waiting to update doas calibration if newer data point appears

ubdbra001:
Fixes #170 

I opted for the second option. Let me know fi you'd like me to add the first option in this PR too.

This also updates the transfer script so that transfers with missing spectra can be simulated. Previously it would have just skipped the images for that timepoint if there was no spectra. It should also be able to handle/simulate cases where one of the image pair is missing too.


twVolc:
Great, thanks. I'll merge this now so that we have this to add to the Merapi computer, but I think breaking out the timeout as a user-defined option in the future might be a nice thing to do too. So perhaps I will keep that issue live and update it with this plan.

Issue #172, opened 2024-07-11
Update "Sequence information" panel when Real-Time Processing


tpering:
As below, this panel should probably update to be the currently processing sequence and range processed:

![image](https://github.com/twVolc/PyCamPermanent/assets/7857899/49e0a58e-3e9a-40c4-b056-cdff46f8cf21)

PR #211, opened 2024-07-31
Add GUI elements for setting real time processing directories

### LINKED ISSUE #189: Add GUI elements for setting watching_dir and FTP_output_dir config options
as title

ubdbra001:
Adds element for setting:
- FTP output directory
- Watching directory

I copied the layout used for setting post processing directories but if you have other ideas about where these should go I'm happy to change it.

closes #189

### DIFF
--- a/pycam/conf/processing_setting_defaults.yml
+++ b/pycam/conf/processing_setting_defaults.yml
@@ -5,7 +5,7 @@
 # General settings
 img_dir: .\pycam\Data\Images\Villarrica\Seq_1\
 dark_img_dir: .\pycam\Data\Images\Villarrica\Cal_1\
-watching_dir: .\pycam\Data\Images\Villarrica\Seq_1\
+transfer_dir: .\pycam\Data\ 
 spec_dir: .\pycam\tests\test_data\test_spectra\
 dark_spec_dir: .\pycam\tests\test_data\test_spectra\
 bg_A_path: .\pycam\Data\Images\Villarrica\Cal_1\2018-03-26T143723_fltrA_1ag_1399829ss_Clear.png
--- a/pycam/doas/doas_worker.py
+++ b/pycam/doas/doas_worker.py
@@ -587,15 +587,15 @@ class DOASWorker(SpecWorker):
         Also starts a processing thread, so that the images which arrive can be processed
         """
         if self.watching:
-            print('Already watching for spectra: {}'.format(self.watching_dir))
+            print('Already watching for spectra: {}'.format(self.transfer_dir))
             print('Please stop watcher before attempting to start new watch. '
                   'This isssue may be caused by having manual acquisitions running alongside continuous watching')
             return
         self.watcher = create_dir_watcher(directory, True, self.directory_watch_handler)
         self.watcher.start()
-        self.watching_dir = directory
+        self.transfer_dir = directory
         self.watching = True
-        print('Watching {} for new spectra'.format(self.watching_dir[-30:]))
+        print('Watching {} for new spectra'.format(self.transfer_dir[-30:]))
 
         # Start the processing thread
         self.start_processing_thread()
--- a/pycam/doas/ifit_worker.py
+++ b/pycam/doas/ifit_worker.py
@@ -848,15 +848,15 @@ class IFitWorker(SpecWorker):
         Also starts a processing thread, so that the images which arrive can be processed
         """
         if self.watching:
-            print('IFit worker: Already watching for spectra: {}'.format(self.watching_dir))
+            print('IFit worker: Already watching for spectra: {}'.format(self.transfer_dir))
             print('IFit worker: Please stop watcher before attempting to start new watch. '
                   'This isssue may be caused by having manual acquisitions running alongside continuous watching')
             return
         self.watcher = create_dir_watcher(directory, recursive, self.directory_watch_handler)
         self.watcher.start()
-        self.watching_dir = directory
+        self.transfer_dir = directory
         self.watching = True
-        print('IFit worker: Watching {} for new spectra'.format(self.watching_dir[-30:]))
+        print('IFit worker: Watching {} for new spectra'.format(self.transfer_dir[-30:]))
 
         # Start the processing thread
         self.start_processing_thread()
@@ -866,7 +866,7 @@ class IFitWorker(SpecWorker):
         if self.watching:
             if self.watcher is not None:
                 self.watcher.stop()
-                print('Stopped watching {} for new images'.format(self.watching_dir[-30:]))
+                print('Stopped watching {} for new images'.format(self.transfer_dir[-30:]))
                 self.watching = False
 
                 # Stop processing thread when we stop watching the directory
--- a/pycam/doas/spec_worker.py
+++ b/pycam/doas/spec_worker.py
@@ -104,7 +104,7 @@ class SpecWorker:
         self.q_spec = queue.Queue()     # Queue where spectra files are placed, for processing herein
         self.q_doas = q_doas
         self.watcher = None
-        self.watching_dir = None
+        self.transfer_dir = None
         self.watching = False
 
         self._dark_dir = None
--- a/pycam/gui/acquisition.py
+++ b/pycam/gui/acquisition.py
@@ -790,8 +790,7 @@ class BasicAcqHandler:
                 return
 
         # Start pyplis worker watcher
-        self.pyplis_worker.start_watching(self.img_dir.root_dir, recursive=True)
-        self.doas_worker.start_watching(self.spec_dir.root_dir, recursive=True)
+        self.pyplis_worker.start_watching_dir()
 
         # Build capture frame
         self.frame = tk.Toplevel()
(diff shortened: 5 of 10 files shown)


twVolc:
Sorry I've been very slow to this, but this look good to me. One thing it makes me think of is that it might be useful to have a label somewhere in the GUI that displays the current watching directory, just so that you can easily look somwhere in the GUI to understand where it's looking for data. My feeling is that it would go in the analysis window within sequence information, perhaps simply updating the the "Sequence directory" label, I think that would make sense right? We could even add  `(Watched)` to the start of the directory path to indicate that it's not a preloaded sequence but it's a directory that's still currently being watched.

tpering:
This also links to issue #172 I think?

ubdbra001:
No worries, I've been much the same over August.
That's a good idea, I have been wondering where to put that info.
So for post-processing the label would stay the same, and for RTP it would change the label to "Watching directory" and display that directory path. Am I interpreting that correctly?

twVolc:
Yes that' what I'm thinking. It might be easier to not actully change the left-hand label at all, and just append `(Watching)` to the end or start of the file path itself so that only that attribute needs accessing/changing. But if it's just as easy to update the left-hand label too then editing that to "Watching directory" sounds fine too. Either/or.

ubdbra001:
I've added an initial update for this that does as you suggest @twVolc.
It should show this directory with (watching) appended when either: the watching directory is selected or 'start watching' is selected (both through the GUI).
I'm not sure if this is ideal though, as it won't change when a config is loaded.

twVolc:
I guess when a config file is loaded it should show the current sequence that's being shown, but if "Start watching" is clicked it should update to the watch directory. But like you say, this isn't ideal as you could load a config file and want to know what watch directory is currently configured, before telling the program to start watching. So perhaps actually we should have a separate watch directory label within that box so that it's clear what it's set up as, then when it's actually watching we maybe have something displayed to show it's now actually watching that directory (e.g. the (watching)) is added at that point. What do you think?

twVolc:
Also unrelated but I notice this PR is now saying "All checks have failed" following #218. Is this working as anticipated or has that been set up slightly incorrectly? It looks a little less appealing not having the big green button saying it can all merge happily.

ubdbra001:
> Also unrelated but I notice this PR is now saying "All checks have failed" following #218. Is this working as anticipated or has that been set up slightly incorrectly? It looks a little less appealing not having the big green button saying it can all merge happily.

This is because I set it up incorrectly #220 should fix it (though getting Github Actions working correctly is a bit of a pain)

twVolc:
> > Also unrelated but I notice this PR is now saying "All checks have failed" following #218. Is this working as anticipated or has that been set up slightly incorrectly? It looks a little less appealing not having the big green button saying it can all merge happily.
> 
> This is because I set it up incorrectly #220 should fix it (though getting Github Actions working correctly is a bit of a pain)

Not sure if my browser just isn't updating but I'm still seeing this problem on this PR after merging #220. Also it's interesting that it's being implemented having only been pulled to dev at this point, not master? I'm guessing these rules just immediately get applied regardless of the branch?

ubdbra001:
It's not you, they are still failing. I've tested the actions locally and they work, so I'm not sure what's going on.
I don't think it's worth spending more time on it until I've got the rest of the PRs sorted. Unless you find the red cross really annoying?

twVolc:
Yeh like you say, not worth wasting time on it now. It's a little annoying not being able to see the green merge telling me things are fine, but I will get over it!

ubdbra001:
> I guess when a config file is loaded it should show the current sequence that's being shown, but if "Start watching" is clicked it should update to the watch directory. But like you say, this isn't ideal as you could load a config file and want to know what watch directory is currently configured, before telling the program to start watching. So perhaps actually we should have a separate watch directory label within that box so that it's clear what it's set up as, then when it's actually watching we maybe have something displayed to show it's now actually watching that directory (e.g. the (watching)) is added at that point. What do you think?

I think a separate watch label is probably the clearest way of relaying the information at a glance. The other alternative would be to put it in the setup paths menu?

twVolc:
Yes setup paths is probably a sensible place to put this. A dedicated label in the Analysis window might be a useful additional place to hold the info but probably isn't critical. Maybe duplicating things isn't best?

ubdbra001:
Okay, in that case I'll put the button to set the value, and the label displaying the current values for both the `FTP_output_dir` and `watching_dir` settings in the setup paths window.
I think it's better to keep to a single place for setting an option, so I'll remove the buttons I added in the menus.

ubdbra001:
I think I have updated everything as discussed:
- You can see the current settings for the FTP output directory and the Watching directory in the processing settings/setup paths menu
- You can also change these values here from the GUI
- The "Sequence directory" label on the Analysis page will update to show "path (watched)" when RTP is selected

I also did a bit of tidying of the code, have a look and make sure it all looks and works okay for you.

ubdbra001:
> I feel like there was a reason why I made them separate, but I can't recall why off the top of my head.

I had a go with this and there's no technical reason why they can't be set by the same config parameter. I think that my thinking in keeping them separate was to make it clear that `FTP_output_dir` is specifically where the files from the Pi get moved to, and they may or may not be watched/processed. To me the `watching_dir` parameter indicates that they will be processed, and doesn't indicate that this is where the files being transferred via FTP will end up.
We might be able to achieve the same thing by renaming these parameters to something better. I'll be the first to admit that my names are not always the most informative.

twVolc:
Yeh I also have a feeling there's a reason why we separated them, but, like you say, maybe it was mainly just to provide clarity on what each path is setting. Overall though, for simplicity of the GUI - not overwhelming the user with options - now that I think more about it it does make sense to keep them as a single parameter. What's your opinion? 

*I don't think your naming conventions are any worse than mine!!

ubdbra001:
I think having them merged is a good idea as it reduces the complexity of the code (only a little but it helps).
But I am unsure if the name `watching_dir` is informative enough to cover both cases.

The only alternative I can think of is `transfer_dir` and we could update the menu items in the RTP menu to explicitly say they are watching the transfer directory? That feels a bit clearer, not a lot but maybe enough?

twVolc:
Yes `transfer_dir` sounds good to me.

ubdbra001:
Okay, all changed.
Try it out and let me know if you're happy with it.

Once you are I'll probably spend a little time sorting out the merge commits and tidying up the git history before we merge.

ubdbra001:
The FTP output is an artefact that I forgot to remove. I'll do that today.

ubdbra001:
Updated to remove references to `FTP_output_dir`, should all be `transfer_dir` now

twVolc:
This looks good to me. It just ahs some conflicts to resolve before merging. I'm never massively confident I fully understand this bit, so do you think this is something you'd be able to do @ubdbra001 ?

ubdbra001:
Okay, give it one last try before you merge, just to make sure I haven't accidentally lost something when sorting out the history and conflicts.
But if you're happy then we're good to merge.

twVolc:
I'm getting an error when I try to run this:

```
Error: Default value for watching_dir missing.

Reverting to config file supplied with PyCam
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\so2_camera_processor.py", line 329, in __init__
    self.load_config(config_path, "default")
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\so2_camera_processor.py", line 352, in load_config
    checked_config = self.check_config_paths(file_path, raw_config)
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\so2_camera_processor.py", line 381, in check_config_paths
    raise ValueError(f"Default value for {path_param} missing.")
ValueError: Default value for watching_dir missing.

During handling of the above exception, another exception occurred:

  File "run_pycam.py", line 1, in <module>
    from pycam.gui.pycam_gui import run_GUI
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\gui\pycam_gui.py", line 14, in <module>
    from pycam.gui.menu import PyMenu
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\gui\menu.py", line 7, in <module>
    from pycam.gui.cfg_menu_frames import geom_settings, process_settings, plume_bg, cell_calib, \
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\gui\cfg_menu_frames.py", line 9, in <module>
    from pycam.gui.figures_analysis import (GeomSettings, ProcessSettings, PlumeBackground, DOASFOVSearchFrame,
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\gui\figures_analysis.py", line 10, in <module>
    from pycam.cfg import pyplis_worker
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\cfg.py", line 16, in <module>
    pyplis_worker = PyplisWorker(config_path=process_settings_path)
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\so2_camera_processor.py", line 342, in __init__
    self.load_config(FileLocator.PROCESS_DEFAULTS, "default")
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\so2_camera_processor.py", line 352, in load_config
    checked_config = self.check_config_paths(file_path, raw_config)
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\so2_camera_processor.py", line 381, in check_config_paths
    raise ValueError(f"Default value for {path_param} missing.")
ValueError: Default value for watching_dir missing.
```

ubdbra001:
Missed a `watching_dir`, should be okay now.

Issue #209, opened 2024-07-30
Interpolated DOAS data when exact data point exists


twVolc:
I'm suddenly seeing a lot of `Interpolated DOAS data for image {}` in RTP (see top screenshot), even though the data for these specific times are available (see bottom screenshot). I'm not 100% certain but I think this has only started appearing since #206. Could this be because data aren't being downloaded in order or another problem? I guess because we are catching the `BaseException` in `update_doas_calibration()` it's hard to know exactly what is going on. For some reason we aren't finding the correct DOAS data point, but seem to find some before and after data points to interpolate. 

It's also odd that it isn't throwing the  `'No DOAS data point within {}s of image time: {}. '` notification, as all DOAS data points other than the correct time will be over the 2s interval used as a limit for interpolation. This must mean the code is finding DOAS data point for the correct time for that image pair, so why are we ending up in that `BaseException` in the first place? 

I'm guessing the #206 code update might raise an exception itself? Perhaps that while-loop doesn't execute properly? 
``` while (datetime.datetime.now() < timeout) and not np.any(img_time < self.doas_worker.results.index):```
@ubdbra001 were you able to check execution of this block? Other than this I can't see why the code has suddenly started going to the BaseException when presumably the DOAS data point is available.

It's probable that we're ending up with the same performance as if we just correctly extracted the DOAS point in the first place, since we will be interpolating to an exact time that exists, but we're executing much more code than we should and it's not the correct execution.

![image](https://github.com/user-attachments/assets/28b84d8c-3417-484a-9213-1a7d86879efd)
![image](https://github.com/user-attachments/assets/076faa22-983b-46e6-8493-ae05198d44b9)

**Comment by twVolc** [comment]
It's not an error with the execution of the while-loop specifically - I've added a print statement to investigate the exception that is being caught it's printing this:
![image](https://github.com/user-attachments/assets/b2b43ba1-85c6-4fd2-a519-228ff4ec6269)
So I'm a bit confused what's happening - it seems the loop is failing to find the `img_time` for a specific spectrum. But it must then be finding it for it to perform the interpolation...

Probably need to investigate what's going on inside the while loop - why isn't this returning a column density:
```
cd = self.doas_worker.results.get(img_time)
if cd is not None: break
time.sleep(0.5)
```
But it seems it used to in the old code.

twVolc:
[...]

PR #217, opened 2024-09-11
Issue 209/new fix

twVolc:
@ubdbra001 if you're able to test this, both in terms of the standard tests to make sure I haven't broken the calculations in some way and in terms of RTP work to see if it seems to run smoothly, that would be great. If I understand correctly, you don't seem to have encountered the #209 error as frequently as we did in Indonesia, so perhaps this isn't going to be a water tight way of checking that it has both removed the issue and not incorporated other annoying isues as before. I think the best we can do is check it on a large amount of data in the RTP mode, to see what happens.

Essentially all I've done is breakout the lock acuqisition into bitsize chunks within `PyplisWorker.update_doas_calibration()` rather than holding the lock for the entirety of the method execution. This should mean that `IFitWorker.results` can be updated whilst `update_doas_calibration()` is in the while-loop check. If I understand correctly there is the smallest chance of a race condition where the check on the `IFitWorker.results` could be updated between the while-loop check on `so2_camera_processor.py` `line 2488` and acquisition of the lock on the next line. But this seems incredibly unlikely to happen, and also still shouldn't lead to a significant bug even if it does happen, as we're still making sure that the lock is acquired throughout assignment of both `cd` and `cd_err`, which I believe is where the issue was arrising.

### DIFF
--- a/pycam/doas/ifit_worker.py
+++ b/pycam/doas/ifit_worker.py
@@ -607,9 +607,10 @@ class IFitWorker(SpecWorker):
         if inplace:
             # Note this function is used in PyplisWorker and self.lock is acquired there, so I don't
             # need to acquire it directly in this function (if we do it would freeze the program.
-            self.results.drop(indices, inplace=True)
-            self.results.fit_errs = fit_errs
-            self.results.ldfs = ldfs
+            with self.lock:
+                self.results.drop(indices, inplace=True)
+                self.results.fit_errs = fit_errs
+                self.results.ldfs = ldfs
             results = None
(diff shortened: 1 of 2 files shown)


ubdbra001:
In my initial tests this looks like a good fix. I'm not getting the "No DOAS data point within 2s of image time..." warning for every image pair. I'm going to keep playing with it to make sure the outputs are consistent.

I don't think I ever hit #203, but I'll have a play around with the lockless code I have to see if I can and then try and recreate it with this new fix.

Issue #141, opened 2024-06-05
No warning if config load fails


twVolc:
At the moment there is no notification to the user if loading a config file fails - the error output is only in the python terminal. It might be best to catch any errors associated with config loading within the main load function and then have an error/warning pop-up in the GUI to say that the load was unsuccessful. 

I think this is most likely to happen when file paths don't exist so certain aspects can't be correctly loaded in - e.g. I'm working a lot with data from an external SSD, and if this isn't plugged in the config load fails and it isn't immediately clear to me why. I wonder if a generic error message saying "Config load has failed, please check all file paths in your config file exist and try again" might help. It would certainly prompt me to remember that I've probably forgotten to plug in my SSD...

tpering:
Does this link to issue #124 as well? If a default config file has been changed to one that has a link to an external on it, the software will fail to load. Is there a way of reverting to the default config file and giving the user a warning?

twVolc:
Ah yes they're very similar! For #124, like you suggest, the software should resort to the default (the stock Villaricca should always be safe and available) and then provide a warning; for this one we just need that warning and the software will stay configured however it was before attempting to load the config.

twVolc:
> for this one we just need that warning and the software will stay configured however it was before attempting to load the config.

Just a note, for this, I guess we first need to save the current config somewhere - maybe to a temp file if we don't want to keep it - then try the config load and if we hit an error we should reload the temporary config file to revert to the current state.

Issue #124, opened 2024-05-10
GUI won't load if there is a load file which points towards an external drive which isn't plugged in


tpering:
As above

twVolc:
Just a note, the software should revert to the default config file in a case like this - the backup default should always be safe and should load. As with #141, a warning should probably then be displayed to indicate that the config file couldn't be loaded. Hopefully should be possible to solve these two simultaneously.

PR #219, opened 2024-09-23
Issue 140/preloaded cal rtp

ubdbra001:
This is an update to PR #193. I rebased the branch on the latest dev commit and resolved the merge conflict, but couldn't push the changes to the open PR (due git history differences).

As discussed in #193 I've added an exception to Pyplisworker and moved the error box to the GUI, which is shown when the exception is raised and caught.

I've also added a warning in the GUI if the user tries to select the preloaded calibration but without specifying a valid file path. It will keep the user there until they provide a valid file or choose a different calibration type.

The last thing that may need resolving is when a user tries to load a config file where `cal_type_int` is 3 but the filename provided doesn't exist. Currently this throws an error but is not handled anywhere. I think it fits in with #124 / #141 so may be better handled within that fix?

### DIFF
--- a/pycam/exceptions.py
+++ b/pycam/exceptions.py
@@ -1,0 +1,2 @@
+class InvalidCalibration(Exception):
+    """Exception highlighing Invalid calibration choice for processing"""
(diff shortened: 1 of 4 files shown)


twVolc:
> The last thing that may need resolving is when a user tries to load a config file where `cal_type_int` is 3 but the filename provided doesn't exist. Currently this throws an error but is not handled anywhere. I think it fits in with #124 / #141 so may be better handled within that fix?

I think that sounds sensible to me. Has this been dealt with in #225 now? Am I right in thinking that this is ready to merge though? I'll do a few checks then merge if so.

ubdbra001:
Yes, checking path existence will be handled in #225, so this eventuality will be covered when that is merged.
I think I'm happy with where this is, so kick the tyres and see if you're also happy with how it handles everything.
Let me know if you encounter anything untoward.

Issue #84, opened 2023-11-30
DOAS config tidying


twVolc:
It might be nice to have a tidier way of dealing with the `IFitWorker` config. My thinking is to have a dictionary within `PyplisWorker` that difines which keys in a config file are associated with `IFitWorker`. Then when `apply_config` is called these files attributes are updated there. Any attributes that need extra work doing (e.g. `ils_path' which needs the ils to actually be loaded from the `ils_path`) should be delt with using properties, so that all of the work under the hood is automatically done when setting the attribute. 

This should tidy things up quite a bit and make it more obvious how to update code when adding new variables to the config file.

PR #221, opened 2024-09-26
Change metadata file paths in config to be relative

### LINKED ISSUE #151: Config file paths might be better as relative
Happy to be told otherwise on this, but I think it would be more suitable to save all of the paths that are to files saved alongside the config file (i.e., `pcs_lines`, `dil_lines`, `image_registration`, `default_cam_geom`) as relative paths; other ones, like `img_dir` should stay as absolute paths. This is just because I sometimes move the processing folder, or rename it, to make things clearer for me, but in doing this I break the config file within it. It makes more sense to me to have that whole folder independent of location - you could move it around but it would still work as long as all the files within the folder remain together.

ubdbra001:
This should close #151 

When the config is saved the metadata files saved alongside it are recorded in the file using relative paths (effectively just filenames)

When a config is loaded the list of path parameters are checked and expanded to absolute paths if needed. Not sure if the list of path params are best at the start of `so2_camera_processor.py` or somewhere else. Happy to move them if you think that makes sense. 

I think it captures all the edge cases around loading and saving paths, but if, during testing, you spot any that I haven't caught do let me know.

### DIFF
--- a/pycam/conf/processing_setting_defaults.yml
+++ b/pycam/conf/processing_setting_defaults.yml
@@ -12,6 +12,7 @@ bg_A_path: .\pycam\Data\Images\Villarrica\Cal_1\2018-03-26T143723_fltrA_1ag_1399
 bg_B_path: .\pycam\Data\Images\Villarrica\Cal_1\2018-03-26T143715_fltrB_1ag_99980ss_Clear.png
 cell_cal_dir: .\pycam\Data\Images\Villarrica\Cal_1\
 cal_type_int: 1
+cal_series_path: ''
 min_cd: 5.0e+16
 plot_iter: 0
 img_buff_size: 1500
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -335,6 +335,25 @@ class PyplisWorker:
         else: 
             [setattr(self, key, value) for key, value in self.config.items()]
 
+    def check_config_paths(self, config_path):
+        config_dir = os.path.dirname(config_path)
+        for path_param in path_params:
+            # Skip if cal_type_int is not pre-loaded
+            if path_param == "cal_series_path" and self.config["cal_type_int"] != 3:
+                continue
+
+            config_value = self.config.get(path_param)
+            # Value could be a string or list of strings, we want to do the same thing to both but iterate over
+            # the list of strings.
+            # Not the most elegent way to do this, but it'll do for now.
+            if type(config_value) is str:
+                new_value = self.expand_config_path(config_value, config_dir)
+                self.config[path_param] = new_value
+            else:
+                for idx, val in enumerate(config_value):
+                    new_value = self.expand_config_path(val, config_dir)
+                    self.config[path_param][idx] = new_value
+
     def expand_config_path(self, path, config_dir):
         """ Converts paths string absolute path if needed"""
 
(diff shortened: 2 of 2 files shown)


twVolc:
This isn't running for me. I'm hitting an error on start-up
```
  File "run_pycam.py", line 1, in <module>
    from pycam.gui.pycam_gui import run_GUI
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\gui\pycam_gui.py", line 14, in <module>
    from pycam.gui.menu import PyMenu
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\gui\menu.py", line 7, in <module>
    from pycam.gui.cfg_menu_frames import geom_settings, process_settings, plume_bg, cell_calib, \
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\gui\cfg_menu_frames.py", line 9, in <module>
    from pycam.gui.figures_analysis import GeomSettings, ProcessSettings, PlumeBackground, DOASFOVSearchFrame, \
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\gui\figures_analysis.py", line 10, in <module>
    from pycam.cfg import pyplis_worker
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\cfg.py", line 16, in <module>
    pyplis_worker = PyplisWorker(config_path=process_settings_path)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\so2_camera_processor.py", line 321, in __init__
    self.load_config(config_path, "default")
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\so2_camera_processor.py", line 332, in load_config
    self.check_config_paths(file_path)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\so2_camera_processor.py", line 354, in check_config_paths
    self.check_path(new_value)
AttributeError: 'PyplisWorker' object has no attribute 'check_path'
```

twVolc:
Also just wondering whether this path checker could be made into its own class? That would solve your query about where to the path_params list. I'm just concious that we've already had discussions about `PyplisWorker` being a monster and this is the kind of processing that feels logical to separate out into its own object (could even be saved in `io_py.py` or similar?).

ubdbra001:
> I'm hitting an error on start-up

Looks like part of a future PR snuck into this code. I've removed it and updated the commit history. Give it another go.

> Also just wondering whether this path checker could be made into its own class? That would solve your query about where to the path_params list. I'm just concious that we've already had discussions about `PyplisWorker` being a monster and this is the kind of processing that feels logical to separate out into its own object (could even be saved in `io_py.py` or similar?).

Excellent idea! I'll have a think about how best to do that while you check you're happy that this works how you want it to.

ubdbra001:
> Shall we wait for this to be incorporated into its own object then merge?

I spent a bit more time thinking about this, and it may be worth thinking bigger and creating a class for handling all aspects of the config files (including loading, checking, saving, etc). It'll make `PyplisWorker` a bit more focussed, and may be a way to implement #84. Currently in places `doas`/`ifit` rely on `PyplisWorker` having the config params, it would be good to try and separate that a bit.
But, that is a bigger bit of work, so may be worth incorporating this (so the functionality is there in the next release) and then coming back to it. What do you think?

twVolc:
Yep I agree with you on this. I'll merge this and open a new issue that links to #84 but is dedicated to breaking out config handling to its own class.

PR #222, opened 2024-09-27
Fix Doas figure spinboxes

ubdbra001:
Spinboxes now execute a command when a value is entered and then focus is moved away (e.g. by clicking on next box).

Try it out and see if you're happy with the change for the wavlength start/end boxes and if so I can add it to other places too.

### DIFF
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -571,6 +571,8 @@ class ImageSO2(LoadSaveProcessingSettings):
                                     command=self.scale_img, font=self.main_gui.main_font)
         self.spin_max.set('{:.2f}'.format(self.tau_max))
         self.spin_max.grid(row=row, column=1, padx=2, pady=2, sticky='ew')
+        self.spin_max.bind('<FocusOut>', self.scale_img)
+        self.spin_max.bind('<Return>', self.scale_img)
         self.auto_tau_check = ttk.Checkbutton(self.frame_opts, text='Auto', variable=self._auto_tau,
                                               command=self.scale_img)
         self.auto_tau_check.grid(row=row, column=2, padx=2, pady=2, sticky='w')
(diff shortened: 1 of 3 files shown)


twVolc:
It would be nice if hitting enter also made the change, rather than needing to click on a different box. But this stuff definitely isn't critical so if it's more than a 5 minute job I wouldn't bother!

Just had a quick look at other places this should perhaps be implemented and there aren't necessarily loads where it's critical. The camera window has an update plot button - I guess this could be removed and changes moved to interactions on the spinboxes. But again this isn't that important. A lot of the spinboxes in the analysis window are most likely to be used as a click rather than writing in them.

ubdbra001:
> It would be nice if hitting enter also made the change

Latest commit should add that, try it and let me know if it's okay.

> Just had a quick look at other places this should perhaps be implemented and there aren't necessarily loads where it's critical. The camera window has an update plot button - I guess this could be removed and changes moved to interactions on the spinboxes. But again this isn't that important. A lot of the spinboxes in the analysis window are most likely to be used as a click rather than writing in them.

Would you be able to add a couple of screenshots highlighting which ones I should update?

twVolc:
These are probably the main places. You could remove the update plot button from the camera window, so this would tidy that up a little. There's also the `shift` button in the DOAS window, that might as well be updated to this too.

change_plot1

change_plot

ubdbra001:
Okay, all done.
Random observation: For the 'Figure options' in the 'Analysis window' would it be worth disabling the spin boxes when the 'Auto' checkbox is clicked? I will admit I spent a good minute changing the values in the spin boxes and being perplexed as to why nothing was changing.

twVolc:
> Okay, all done. Random observation: For the 'Figure options' in the 'Analysis window' would it be worth disabling the spin boxes when the 'Auto' checkbox is clicked? I will admit I spent a good minute changing the values in the spin boxes and being perplexed as to why nothing was changing.

Haha yes I think it would!!

ubdbra001:
Done, so I think this is ready to merge. Unless you've spotted something else related that could be fixed here?

twVolc:
This is all excellent! It makes quite a big difference to interactivity

PR #225, opened 2024-10-02
Provide feedback when config loading fails

### LINKED ISSUE #141: No warning if config load fails

ubdbra001:
Fixes #141

This makes sure that the paths specified in the config file actually exist. If not there are two different behaviours depending on when the config file is being loaded:
1. If it is the when PyCam starts (i.e. the default config) then it will fall back to the config file that is supplied with PyCam, and show a warning box once the GUI has loaded (it will also print a warning in the console).
2. If it is when a user config is loaded, then it will display an error box and will keep the currently loaded config. If running without a GUI it will just throw an exception.

The actual messages may need some work.

### DIFF
--- a/pycam/gui/menu.py
+++ b/pycam/gui/menu.py
@@ -799,6 +799,11 @@ class LoadFrame(LoadSaveProcessingSettings):
             cfg.current_dir_img.root_dir = ftp_output_dir
             cfg.current_dir_spec.root_dir = ftp_output_dir
 
+        if pyplis_worker.missing_path_param_warn is not None:
+            messagebox.showwarning("Missing path params not updated",
+                                   pyplis_worker.missing_path_param_warn)
+            pyplis_worker.missing_path_param_warn = None
+
     def reset_pcs_lines(self):
         """Reset current PCS lines"""
 
--- a/pycam/gui/menu.py
+++ b/pycam/gui/menu.py
@@ -25,6 +25,7 @@ import tkinter.ttk as ttk
 from tkinter import filedialog
 from tkinter import messagebox
 from shutil import copyfile
+from inspect import cleandoc
 import os
 import threading
 
@@ -755,7 +756,18 @@ class LoadFrame(LoadSaveProcessingSettings):
                 initialdir=self.init_dir)
         
         if len(filename) > 0:
-            self.pyplis_worker.load_config(filename, "user")
+            try:
+                self.pyplis_worker.load_config(filename, "user")
+            except FileNotFoundError as e:
+                error_msg = cleandoc(
+                    f"""
+                    Config load unsuccessful\n
+                    Error: {e}
+                    """)
+
+                messagebox.showerror("Config load failure",
+                                     error_msg)
+                return
 
             self.reload_config()
 
--- a/pycam/gui/pycam_gui.py
+++ b/pycam/gui/pycam_gui.py
@@ -134,6 +134,11 @@ class PyCam(ttk.Frame):
             cfg.current_dir_img.root_dir = ftp_output_dir
             cfg.current_dir_spec.root_dir = ftp_output_dir
 
+        if pyplis_worker.load_default_conf_errors is not None:
+            messagebox.showwarning("Default Config Error",
+                                   pyplis_worker.load_default_conf_errors)
+            pyplis_worker.load_default_conf_errors = None
+
     def exit_app(self):
         """Closes application"""
         if messagebox.askokcancel("Quit", "Are you sure you want to quit?"):
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -319,6 +319,7 @@ class PyplisWorker:
 
         self.geom_dict = {}
 
+        self.missing_path_param_warn = None
         self.config = {}
         self.raw_configs = {}
         self.load_config(config_path, "default")
@@ -343,13 +344,25 @@ class PyplisWorker:
             [setattr(self, key, value) for key, value in self.config.items()]
 
     def check_config_paths(self, config_path, raw_config):
+        missing_path_params = []
         config_dir = os.path.dirname(config_path)
+
         for path_param in path_params:
             # Skip if cal_type_int is not pre-loaded
             if path_param == "cal_series_path" and raw_config["cal_type_int"] != 3:
                 continue
 
             config_value = raw_config.get(path_param)
+
+            # If there is no value in the config for this parameter then record and skip
+            # Unless it is when no other config file has been loaded, then throw an error.
+            if config_value is None:
+                if self.config:
+                    missing_path_params.append(path_param)
+                    continue
+                else:
+                    raise ValueError(f"Default value for {path_param} missing.")
+
             # Value could be a string or list of strings, we want to do the same thing to both
             # but iterate over the list of strings.
             # Not the most elegent way to do this, but it'll do for now.
@@ -360,6 +373,16 @@ class PyplisWorker:
                 for idx, val in enumerate(config_value):
                     new_value = self.expand_check_path(val, config_dir)
                     raw_config[path_param][idx] = new_value
+
+        if missing_path_params:
+            miss_param_str = [f"- {par}" for par in missing_path_params]
+
+            self.missing_path_param_warn = "\n".join(
+                ["The following parameters were not present in the loaded config:",
+                 *miss_param_str,
+                 "Current values were retained."])
+            print(self.missing_path_param_warn)
+
         return raw_config
 
     def expand_check_path(self, value, dir, path_param):
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -369,6 +369,9 @@ class PyplisWorker:
         # (specified by a path relative to pycam location)
         if not os.path.isabs(config_dir):
             return path
+        
+        if path == '':
+            raise FileNotFoundError("Path not specified")
         # If it's an absolute path then just use as is
         elif os.path.isabs(path):
             return path
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -329,10 +329,11 @@ class PyplisWorker:
 
         file_path = os.path.normpath(file_path)
         with open(file_path, "r") as file:
-            self.raw_configs[conf_name] = yaml.load(file)
+            raw_config = yaml.load(file)
 
+        checked_config = self.check_config_paths(file_path, raw_config)
+        self.raw_configs[conf_name] = checked_config
         self.config.update(self.raw_configs[conf_name])
-        self.check_config_paths(file_path)
 
     def apply_config(self, subset = None):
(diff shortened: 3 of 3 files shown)


ubdbra001:
Should also fix #124

twVolc:
I'm getting issues with this on both config files that load in dev and config files that shouldn't load. They both throw the same error:
```
Exception in Tkinter callback
    return self.func(*args)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\gui\menu.py", line 739, in load_config_file
    self.pyplis_worker.load_config(filename, "user")
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\so2_camera_processor.py", line 341, in load_config
    checked_config = self.check_config_paths(file_path, raw_config)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\so2_camera_processor.py", line 368, in check_config_paths
    for idx, val in enumerate(config_value):
TypeError: 'NoneType' object is not iterable
```
Probably the easiest way to reproduce this error is by editing the default file - I've just edited `bg_B_path` so that it doesn't exist. It throws this error but doesn't create a pop-up in the GUI. This error is identical when I try to load some of my config files that will load in dev. Note I've had to change the file extension to .txt as I couldn't attach .yml here.

[processing_setting_defaults - Copy.txt](https://github.com/user-attachments/files/17228900/processing_setting_defaults.-.Copy.txt)

ubdbra001:
That's odd, the file you provided doesn't throw the error for me.

I've added a handler for when getting the config value fails (which is why it's trying to iterate over a `None`), could you have another go?

twVolc:
That seems to have caught that error. Oddly my other config files has progressed to a new error, which I'm guessing is related to nnew values that have been added to the config since I generated this file. Just thinking some backwards compatibility might be nice, if there's a way of dealing with missing keys, or just specific keys that have been added recently?
```
Exception in Tkinter callback
    return self.func(*args)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\gui\menu.py", line 739, in load_config_file
    self.pyplis_worker.load_config(filename, "user")
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\so2_camera_processor.py", line 341, in load_config
    checked_config = self.check_config_paths(file_path, raw_config)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamTests\PyCamPermanent\pycam\so2_camera_processor.py", line 362, in check_config_paths
    raise ValueError(f"{path_param} is None")
ValueError: watching_dir is None
```

I got it to load by adding `watching_dir: ./` to the config, but I'm just wondering whether it would make more sense to have keys stay as default or as their current settings if they're missing from the config file? In general this might be a bit of an edge case, but for the backwards compatibility with `watching_dir` I think we should at least do something about that.

ubdbra001:
Yep, I agree.

It should actually do that anyway: The values that appear in the loaded config should replace the values in the already loaded config, but if one doesn't appear in the new config the default value should be retained. In this case I don't think I planned for the fact some may be missing.

I'll replace the exception with a warning message in the console (not sure how I'd do it in the GUI, but could try) and continue to the next `path_param` to avoid this.

twVolc:
> I'll replace the exception with a warning message in the console (not sure how I'd do it in the GUI, but could try) and continue to the next `path_param` to avoid this.

To do it in the GUI I think you would need to have some kind of list accumulating all of the missing keys when iterating through them, then you can add those to a `messagebox.showwarning` at the end, which would list all of them in one widget. I don't know where exactly in the code that list would be accumulated though.

ubdbra001:
Could do it in the config checking method?
Then we could have it print out a single warning and the list in the console at the end of the check (which makes things a bit less cluttered there) and pass the same list back to the GUI for a warning box.

I can play around with doing that if you think it's worth it for this PR.

twVolc:
I think it fits under this PR of providing feedback during config loading, so let's hold off merging for now. If that turns into a horribly big piece of work don't worry about it though - let me know and I'll just merge as is.

ubdbra001:
It'll now show a message in the console, and a warning after loading detailing the config path parameters that weren't in a loaded config file.
Give it a go and see what you think.

twVolc:
I might be missing things here but I'm not sure I'm getting the anticipated results. I've tried commenting out some of the lines and they seem to lead to different results depending on what line it is. It might just be that this is quite complex to get all options working in the same way if some are missing? Are there only specific ones that are set to be recorded if they're missing? A few I've tried commenting out are:

1. `img_dir`. This then raises a dialogue box after the GUI has loaded, asking you to select an image sequence. This works quite well actually.
2. 'dark_img_dir` doesn't notify the user, but you can see when you got to setup paths that it hasn't been loaded.
3. `bg_path..` fails to load with 
```Traceback (most recent call last):
  File "run_pycam.py", line 3, in <module>
    run_GUI()
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\gui\pycam_gui.py", line 163, in run_GUI
    myGUI = PyCam(root, x_size, y_size)
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\gui\pycam_gui.py", line 94, in __init__
    self.info_load()
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\gui\pycam_gui.py", line 107, in info_load
    process_settings.initiate_variables(self)
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\gui\figures_analysis.py", line 2926, in initiate_variables
    self.load_defaults()
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\gui\misc.py", line 305, in load_defaults
    self.gather_vars()
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\gui\figures_analysis.py", line 3333, in gather_vars
    pyplis_worker.load_BG_img(self.bg_A_path, band='A')
  File "C:\Users\thom_\Documents\Work\Python\pycam_dev\PyCamPermanent\pycam\so2_camera_processor.py", line 1070, in load_BG_img
    raise ValueError('File path specified for background image does not exist: {}'.format(bg_path))
ValueError: File path specified for background image does not exist:
```

I might be misunderstanding what has been done here, so let me know what you think to this - is this planned or a bug? Is it more just associated with the files that are in the config path fodler spcifically that are dealt with here? I guess in realtiy the only attributes that would need to be caught are ones that might not be present in older version, to ensure backwards compatibility - beyond that maybe it's fine to assume parameters will exist. I've tried `watching_dir` which is one that might be missing from older configs, and that doesn't seem to raise a warning when it's commented out.

ubdbra001:
Just to clarify: this is when you have loaded the GUI and you are trying to load in a new config, right?

twVolc:
Ah no sorry - that's probably where I'm getting confused! This was if using this file as the start-up config file

ubdbra001:
That's fine, just need to know what you're doing to try and replicate it.

twVolc:
Just tested things in they way they're supposed to....

This looks good. A few comments that hopefully help:

- The warning says "Default values were retained". Should this be "Current values were retained"? I can't remember what we decided on this but think it was that the current state of the software remains unchanged for attributes that can't be found? So default might be a bit misleading.

- This might be pedantic and overkill, but a little like the warning when an attribute is missing, it could be nice to have the Error message state what attribute/filename is wrong when you get the "Config load failure" box. This would direct the user to the error. Unlike the warning box, for simplicity this could just be the first error that is hit, rather than complicating things with having to check through the entire config file to list all the ones that don't work. This is only a nice-to-have though, there's other thigns to prioritise if this addition would take a long time.

twVolc:
> Ah no sorry - that's probably where I'm getting confused! This was if using this file as the start-up config file

Coming back to this, these acutally probably do still need handling - I guess they should revert to default? Or it could revert to default, then load the new config on top, so that it throws the warning box with all the parameters that haven't been loaded - just like what happens when you use the GUI to load a new config.

ubdbra001:
> Coming back to this, these acutally probably do still need handling - I guess they should revert to default? Or it could revert to default, then load the new config on top, so that it throws the warning box with all the parameters that haven't been loaded - just like what happens when you use the GUI to load a new config.

Yeah, currently it has two different failure modes:
1. If the default config has missing paths then it should fall back to the supplied config and show an error
2. If a user loaded config has missing paths then it should retain the default paths and tell the user which paths/config params are causing the issues

Mode 1 has only accounted for the paths being wrong, not missing entirely and so is getting caught by mode 2 and is producing the different behaviour you see

ubdbra001:
> The warning says "Default values were retained". Should this be "Current values were retained"? I can't remember what we decided on this but think it was that the current state of the software remains unchanged for attributes that can't be found? So default might be a bit misleading.

You're right we should be consistent with this. I don't recall what we decided either (I don't think I implemented it), but at the moment it does write over current settings rather than defaults. Shall we take a decision on this once and for all and then we can get the message to match up.

> it could be nice to have the Error message state what attribute/filename is wrong when you get the "Config load failure" box

That should be pretty easy to do. the info is already included in the error when it is raised and it should be easy to pass that to the message box. Would it be helpful for all of the warning/error boxes to have some indication of what parameter caused the issue?

twVolc:
> You're right we should be consistent with this. I don't recall what we decided either (I don't think I implemented it), but at the moment it does write over current settings rather than defaults. Shall we take a decision on this once and for all and then we can get the message to match up.

I think how you have it now is good - keeping the settings as they currently are if there isn't any new info to load- I don't think it makes sense to revert to default in these scenarios. So in this case all that needs changing is that wording from Default to Current.

> That should be pretty easy to do. the info is already included in the error when it is raised and it should be easy to pass that to the message box. Would it be helpful for all of the warning/error boxes to have some indication of what parameter caused the issue?

I think how the warning box is currently setup - with all missing parameters listed - is good for that warning case. For the Error boxes I think just the name of the attribute that caused the error would be sufficient.

ubdbra001:
Okay, I think I covered everything in the latest updates.
Let me know what you think.

twVolc:
Yes this is excellent, the messages are informative and it's really handy to be able to start-up the software even when there's an issue with the config file. 

Feel free to merge yourself when you're happy with the commit history.

PR #226, opened 2024-10-02
Update readme to move away from Anaconda installer

### LINKED ISSUE #215: Move from Anaconda?
There was some mention of this in the RSE/ITS slack a little while ago, but it looks like Anaconda are clamping down on what it sees as usage outside its ToS: [Article](https://www.theregister.com/2024/08/08/anaconda_puts_the_squeeze_on/)

I've not heard anything about the University's position on this and I'm not sure if the PyCam code would violate those terms, but it may be worth thinking about a move from it regardless. Given we use pip for installing pretty much all the dependencies we should be okay on that front (we should be able to use `venv` or something similar). So, I think the only thing that may present an issue is getting the right version of Python.

ubdbra001:
closes #215 

More info can be found in the issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,36 +1,47 @@
 # PyCam project repo
+
 ## Documentation
-- Detailed user and quick start guides can be found at: https://www.volcanotech.org/software.html 
+
+Detailed user and quick start guides can be found at [volcanotech.org](https://www.volcanotech.org/software.html)  
 
 ## Pre-requsities
 
 - Windows 10/11
-- `conda` installed ([installation instructions](https://docs.conda.io/projects/conda/en/latest/user-guide/install/windows.html))
+- `conda` installed ([Download page and installation instructions](https://conda-forge.org/download/))
 - `git` installed (optional but recommended; [installation instructions](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git))
+
+### Note
+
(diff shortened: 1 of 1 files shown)


twVolc:
Sorry I'm a little confused on this (although maybe I just don't need to understand it all...). Is this `conda` somehow different to the `conda` we currently use that has licencing issues. Is this `conda` the same as `mamba` or are they all different?

ubdbra001:
As far as I understand it: the `conda` software is the same, as this is open source software.

But the one included in `mini-forge` uses `conda-forge` channel/repos to get Python and packages, this is owned and maintained by the community and so is free for everyone to use (and will remain so).
The `conda` from Anaconda points to the Anaconda repos, and a subject to their terms (and any changes they make to them, which is what caused the uproar in the first place).

`mamba` is a reimplementation of `conda`  that was built to be faster. Although I think part of it got integrated into `conda`.

An alternative to switching would be to add a line in the instructions that tells the Anaconda version of `conda` to only use the `conda-forge` channels, if you think that would be easier?

twVolc:
Great, thanks that makes sense (I think). This seems like a straightforward solution so lets stick to this - the conda-forge installation page seems straightforward. I don't particularly want to have to check installing the new version of conda now but it sounds like you've checked this and itruns smoothly? So I'm happy to merge if you are

PR #228, opened 2024-10-02
Add points for Analysis time series plots

### LINKED ISSUE #214: Emission rate plotting doesn't show single points
I'm pretty sure I removed markers from the emission rate plot because it made things look messy, but I've realised that using just lines means that when there are individual data points they don't show on the plot - this is relevant for flow_histo and flow_hybrid which can regularly output NaNs. If a data point is surrounded by 2 NaNs, i.e. there aren't 2 consecutive numeric points, the vaues won't appear on the plot. I think the solution is to use very small markers, which means that they don't make the plot messy (they blend in with the size of the lines) but they appear as single points when lines aren't drawn.

ubdbra001:
Fixes #214

### DIFF
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -1126,6 +1126,8 @@ class TimeSeriesFigure:
                             }
         self.colours = self.pyplis_worker.fig_tau.line_colours
         self.marker = '.'
+        self.ER_markersize = 3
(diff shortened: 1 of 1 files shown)


twVolc:
Happy for this to be made an open PR now. I think what you've done looks good and solves the issue. There might come a time way down the line where we do a complete overhaul of plotting design, but for the #214 bug this does a good job.

Issue #224, opened 2024-10-02
Create a config handler class


twVolc:
As discussed in PR #221, and also relating to #84 somewhat, it would make sense to move all config handling to its own class (`Config`?). This would deal with all config file I/O work and setting up both `PyplisWorker` and `IFitWorker` correctly. this would tidy up `PyplisWorker` quite a bit, ensuring that it stays focussed on data processing and so is slightly less cluttered.

ubdbra001:
Also relates to #67

twVolc:
> * The GUI reads config values when initialised and when a new config is loaded, and writes to the config when something is changed in the GUI. The GUI will need access to the config to set the initial values for most GUI elements.

I might be missing something here, but would it be possible to have the `pyplis_worker`/`doas_worker` writing updates to the config file, rather than the GUI? That way maybe it's possible to have the GUI itself not really interacting with the config file and it's just facilitiating interactions between `pyplis_worker`/`doas_worker` and the config. That way maybe we avoid the issue of tangling too much into the GUI that doesn't need to be - in other words, if we have the GUI interacting too much with the `ConfigHandler` might we end up with a similar situation to what we have now with `PyplisWorker` and the GUI, where we wish they weren't so tangled and are now having to work to disentangle?

It's probably that I'm just struggling to picture the plans exactly - might be easier over an online discussion or with diagrams? In the end though, I'm happy to go with whatever you think you'd be able to get working in the time available (and perhaps in the shortest amount of time).

ubdbra001:
There does need to be interaction between the GUI and config as the GUI uses the values in the config file to populate itself, and, when saving, the config should reflect the state of the GUI.

It will still be going through `pyplis/doas_worker` as these will have an instance of the `ConfigHandler` class as an attribute, and then when it comes to saving to a file either will be able to call the saving method in that instance.
We could have it so that the GUI reads and writes its state to the attributes in the `pyplis/doas_worker` instances, and then the `ConfigHandler` class pulls it from there, but that feels like an un-needed step when its possible to access the `ConfigHandler` instance directly through each.

> if we have the GUI interacting too much with the `ConfigHandler` might we end up with a similar situation to what we have now with `PyplisWorker` and the GUI, where we wish they weren't so tangled and are now having to work to disentangle?

[...]

PR #230, opened 2024-10-02
Move wavelength doas parameters to config file

### LINKED ISSUE #191: Making all doas/ifit parameters available to config
I think we've had a few discussions about this but I can't see any concrete issue that specifically addresses this.

Essentially, there are some parameters that can be adjusted in the GUI but that are hard-coded rather than defined by the config file initially. Most importantly that I can think of:
-`start_fit_wave`, `end_fit_wave`
-`start_stray_wave`, `end_stray_wave`
-`shift`
There may be others but these are some important ones that can quite often be adjusted by the user. It would be really useful if these can be added to the config file.

#67 might be a useful reference and is at least somewhat related. #84 might also be a bit relevant here.

ubdbra001:
Fixes #191

### DIFF
--- a/pycam/conf/processing_setting_defaults.yml
+++ b/pycam/conf/processing_setting_defaults.yml
@@ -68,6 +68,11 @@ fov_rad: 0
 centre_pix_x: 0
 centre_pix_y: 0
 doas_cal_adjust_offset: 0
+start_stray_wave: 293
+end_stray_wave: 296
+start_fit_wave: 308
+end_fit_wave: 318
+shift: 0
 
 # Plume speed algorithms
 flow_raw: 1
--- a/pycam/conf/processing_setting_defaults.yml
+++ b/pycam/conf/processing_setting_defaults.yml
@@ -73,7 +73,7 @@ end_stray_wave: 296
 start_fit_wave: 308
 end_fit_wave: 318
 start_fit_wave_ld: 312
-start_fit_wave_ld: 322
+end_fit_wave_ld: 322
 shift: 0
 
 # Plume speed algorithms
--- a/pycam/conf/processing_setting_defaults.yml
+++ b/pycam/conf/processing_setting_defaults.yml
@@ -72,6 +72,8 @@ start_stray_wave: 293
 end_stray_wave: 296
 start_fit_wave: 308
 end_fit_wave: 318
+start_fit_wave_ld: 312
+start_fit_wave_ld: 322
 shift: 0
 
 # Plume speed algorithms
--- a/pycam/doas/doas_worker.py
+++ b/pycam/doas/doas_worker.py
@@ -48,11 +48,6 @@ class DOASWorker(SpecWorker):
         # Initial Definitions
         # ======================================================================================================================
         self.ppmm_conversion = 2.7e15   # convert absorption cross-section in cm2/molecule to ppm.m (MAY NEED TO CHANGE THIS TO A DICTIONARY AS THE CONVERSION MAY DIFFER FOR EACH SPECIES?)
-        
-        self._start_stray_wave = 280    # Wavelength space stray light window definitions
-        self._end_stray_wave = 290
-        self._start_fit_wave = 305  # Wavelength space fitting window definitions
-        self._end_fit_wave = 320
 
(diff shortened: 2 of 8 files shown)


twVolc:
Stray and Fit wavelengths work well but the shift parameter doesn't seem to load correctly (the GUI value isn't updated and you can see from the spectrum that a shift hasn't been applied).

ubdbra001:
And here I was wondering why this was still a draft. Will add that now.

ubdbra001:
In the part of the DOAS window with the shift spin box, are the x-axis values meant to update when you change the shift value?

twVolc:
No they will stay the same as you're technically still fitting in the same window, you're just shifting the measured spectrum relative to the reference spectrum (it's essentially used when the spectrometer calibration isn't quite right, e.g. thermal shifting).

ubdbra001:
Two more questions:
1. Do you want to add `shift_tol` to the config file?
2. Did you decide how you wanted to proceed with the `start/end_fit_wave_ld` parameters?

twVolc:
> 1. Do you want to add `shift_tol` to the config file?

I think there's no reasonable likelihood that I get the `DOASWorker` object properly working any time soon (unless I suddenly find a lot of free time!), so I think `shift_tol` can just stay out of the config for now. What might be useful is if there's some documentation somewhere (perhaps just comments in the code, or maybe there's a more sensible place?) that details the typical procedure for incorporating new attributes into the config file - i.e. places that need to be updated. I'm guessing this will be case-by-case specific, as different aspects of the GUI may need updates too, but a general one might help future development. This could be a little like `NadeauFlowSettings` documentation at the start of the class. I guess this could link to #224, it might be where these kinds of details are held, if at all.

> 2. Did you decide how you wanted to proceed with the `start/end_fit_wave_ld` parameters?
 I guess this is `start/end_fit_wave_2`? There's no harm in putting these in the config if it's straightforward to do so. I think in reality  quite a bit of work needs to be done on the LD side of the spectrometer, but this isn't something I'm able to prioritise so for the time being just adding this to the config is probably a fine start.

ubdbra001:
> I guess this could link to https://github.com/twVolc/PyCamPermanent/issues/224, it might be where these kinds of details are held, if at all.

I was thinking the same thing, though having a class to handle all the config stuff should remove the need to add this kind of documentation (at least for the backend), i.e. adding something new in the config should mean it automatically gets added to the correct backend worker. This is how it works currently for `pyplis_worker` but for `doas_worker` it's a bit trickier.

 The frontend is a bit more tricky and element dependent, but I'm happy to add what I can.

> I guess this is `start/end_fit_wave_2`? There's no harm in putting these in the config if it's straightforward to do so. I think in reality quite a bit of work needs to be done on the LD side of the spectrometer, but this isn't something I'm able to prioritise so for the time being just adding this to the config is probably a fine start.

Okay, will do.

twVolc:
> Stray and Fit wavelengths work well but the shift parameter doesn't seem to load correctly (the GUI value isn't updated and you can see from the spectrum that a shift hasn't been applied).

Just working back through PRs to see which are ready to go. Looks like this one is still waiting for this update so I won't merge yet.

ubdbra001:
Okay, I've added the `start/end_fit_wave_ld` as far as I could tell this is just loaded into `DoasWorker` but there isn't anything that corresponds to it in the GUI. Let me know if I've got that wrong.

I'll leave `shift_tol` until I address #224.

twVolc:
No this is correct. This LD work is something I haven't fully brought into the GUI yet. Probably at some point I should make this an option but it's nowhere near a priority as we don't use this much and I think really we're very unlikely to change these values. Having it in the config is nice though, for easy access if we do ever want to change it.

I'll do a quick test then aim to merge.

> I'll leave `shift_tol` until I address #224.

This sounds like a good plan.

pycam/conf/processing_setting_defaults.yml
```diff
@@ -68,6 +68,13 @@ fov_rad: 0
 centre_pix_x: 0
 centre_pix_y: 0
 doas_cal_adjust_offset: 0
+start_stray_wave: 293
+end_stray_wave: 296
+start_fit_wave: 308
+end_fit_wave: 318
+start_fit_wave_ld: 312
+start_fit_wave_ld: 322
```
twVolc, line 76: This duplicate needs changing to `end_fit_wave_ld` as this config currently means a yaml error is thrown when trying to strat the software.

Issue #237, opened 2024-10-15
Duplicate times in saved calibration file - can't load the calibration file


twVolc:
I'm not quite sure how this has happened, whether it unique to the data I'm working with or whether this is a new bug caused by some code update. My `full_calibration.csv` file for a latest bit of processing has lots of duplicate times in the file, which is leading to an error when trying to load the calibration. It hits the error in `line 2431` (in dev) of `so2_camera_processor.py`:
           
` closest_index = self.calibration_series.index.get_indexer([self.img_A.meta['start_acq']], method='nearest')`

Throws the error:
```
Traceback (most recent call last):
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\run_pycam.py", line 3, in <module>
    run_GUI()
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\pycam\gui\pycam_gui.py", line 159, in run_GUI
    myGUI = PyCam(root, x_size, y_size)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\pycam\gui\pycam_gui.py", line 94, in __init__
    self.info_load()
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\pycam\gui\pycam_gui.py", line 127, in info_load
    pyplis_worker.load_sequence(pyplis_worker.img_dir, plot_bg=False)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\pycam\so2_camera_processor.py", line 968, in load_sequence
    self.process_pair(self.img_dir + '\\' + self.img_list[0][0],
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\pycam\so2_camera_processor.py", line 3539, in process_pair
    self.generate_optical_depth(plot=plot_img, plot_bg=plot_bg, run_cal=force_cal, img_path_A=img_path_A,
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\pycam\so2_camera_processor.py", line 2312, in generate_optical_depth
    self.img_cal = self.calibrate_image(self.img_tau, run_cal_doas=run_cal)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\pycam\so2_camera_processor.py", line 2369, in calibrate_image
[...]

PR #238, opened 2024-10-15
Remove duplicate calibration points

twVolc:
Allows loading of calibration files with duplicate times by removing the duplicate lines.


twVolc:
@ubdbra001 shall I close this PR? I just checked through #237 and from what I understand I think you've found and fixed this issue as part of #240? So perhaps it's not necessary to have this duplicate check. Or if you think it might be good to add this extra safety net that removes duplicates I can merge it? These duplicate data points were a breaking bug (couldn't load the calibration at all), so want to make sure we definitely avoid it.

ubdbra001:
I think it would be good to just close this. I think while the extra safety net may be useful it is likely to mask things not working properly when the file is generated.

PR #240, opened 2024-10-24
Add blank data for missing spectra timepoints

### LINKED ISSUE #237: Duplicate times in saved calibration file - can't load the calibration file

ubdbra001:
This fixes #237

When there is an image pair without an associated spectra timepoint the calibration file will include blank columns for the image/spectra data (tau, cd, etc) and reuse the last calibration values (coeff, MSE, etc).

It will only do this once the calibration has done it's initial run, prior to this it will just drop the missing timepoints. This is to maintain consistency between the "fov find" and "fov fixed" approaches.

Emission results are consistent with dev branch results.

### DIFF
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -2611,11 +2611,11 @@ class PyplisWorker:
                     # point for img_time, so we can raise the error which moves us onto the interpolation section.
                     with self.doas_worker.lock:
                         cd = self.doas_worker.results.get(img_time)
+                        if cd is None:
+                            raise KeyError(f"spectra for {img_time} not found")
                         # Get index for cd_err
                         cd_err = self.doas_worker.results.fit_errs[
                             np.where(self.doas_worker.results.index.array == img_time)[0][0]]
-                    if cd is None:
-                        raise KeyError(f"spectra for {img_time} not found")
 
             except BaseException as e:
                 with self.doas_worker.lock:
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -2618,6 +2618,11 @@ class PyplisWorker:
                             np.where(self.doas_worker.results.index.array == img_time)[0][0]]
 
             except BaseException as e:
+                
+                # Give warning when unexpected (i.e. non-KeyError) is caught.
+                if type(e) is not KeyError:
+                    print(f"Unexpected error: {e}" )
+
                 with self.doas_worker.lock:
                     # If there is no data for the specific time of the image we will have to interpolate
                     dts = self.doas_worker.results.index - img_time
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -2629,6 +2629,20 @@ class PyplisWorker:
                         print('No DOAS data point within {}s of image time: {}. '
                               'Image is not added to DOAS calibration'.format(self.max_doas_cam_dif,
                                                                               img_time.strftime('%H:%M:%S')))
+
+                        # Only add if the calibration is already available
+                        if self.fit_data.size > 0:
+
+                            # Add empty tau_values
+                            tau_val_ncols = self.tau_vals.shape[1]
+                            empty_tau_vals = np.column_stack([img_time, *np.repeat(np.nan, tau_val_ncols-1)])
+                            self.tau_vals = np.append( self.tau_vals, empty_tau_vals, axis = 0)
+
+                            # Add the last calibration values
+                            last_cal = self.fit_data[-1][1:]
+                            fit_data = np.hstack((img_time, *last_cal))
+                            self.fit_data = np.append(self.fit_data, fit_data[np.newaxis, :], axis = 0)
+
                         return
 
                     zero = datetime.timedelta(0)


pycam/so2_camera_processor.py
```diff
@@ -2611,11 +2611,11 @@ def update_doas_calibration(self, img_tau=None, force_fov_cal=False):
                     # point for img_time, so we can raise the error which moves us onto the interpolation section.
                     with self.doas_worker.lock:
                         cd = self.doas_worker.results.get(img_time)
+                        if cd is None:
+                            raise KeyError(f"spectra for {img_time} not found")
                         # Get index for cd_err
                         cd_err = self.doas_worker.results.fit_errs[
                             np.where(self.doas_worker.results.index.array == img_time)[0][0]]
-                    if cd is None:
```
twVolc, line 2617: I'm slightly confused as to why I put this here, but I have a feeling it might be for a reason. I think it's to free up the lock for a brief second, which means that the IFitWorker thread can access results and add any new data. Meaning that once we enter the Exception there might be new data in there. Is there a specific reason on your end that it has to be moved back inside the lock?

ubdbra001: Basically to catch a missing spectra time point as clearly/cleanly as possible. With the way it was originally, with a missing time point: cd was `None` but it wasn't getting to the `KeyError` as the next line throws an error:
```python
cd_err = self.doas_worker.results.fit_errs[np.where(self.doas_worker.results.index.array == img_time)[0][0]]
```
and so I wanted the behaviour to be a bit more clearly defined.
To be honest, we should probably change the exception to only catch `KeyError` otherwise anything that goes wrong in this block (even it is unrelated) will just get caught and ignored.

twVolc: That makes sense, not sure what I was thinking when I wrote it this way.... 

I'm just thinking how to deal with this issue best - from memory, we need both attempts at retrieving `cd` and `cd_err` to be within the same lock, otherwise we could encounter the bug the locks were aiming to avoid (I think!). Although looking again at it, perhaps it's just that the entire `cd_err` line needs to have the lock acquired otherwise aspects of it could execute, but not fully, which is when we hit the race condition. So maybe rather than moving the `KeyError` catch inside that initial lock, we keep it outside, but then move the `cd_err` line after this, with it's own acquisition of the lock prior to executing. So the code would look like this:
```
with self.doas_worker.lock:
    cd = self.doas_worker.results.get(img_time)
if cd is None:
    raise KeyError(f"spectra for {img_time} not found")
# Get cd_err
with self.doas_worker.lock:
    cd_err = self.doas_worker.results.fit_errs[
            np.where(self.doas_worker.results.index.array == img_time)[0][0]]
```

How does this look? This lock work is a bit of a pain and probably unneccessarily convoluted, but for the time-being it's seemingly hard to extensively test these things without having a camera to connect to as it seems that your simulated data transfer doesn't cause the same issue?

I can't remember if there's any other errors that can get thrown in this section of code, but worry there might be others which are why I caught the `BaseException`. I understand this isn't a great idea, I should have worked out which ones I want to catch and stated that. I worry we might break some functionality if we move away from `BaseException` without extensive testing to see if there's any other errors beyond `KeyError` that should be caught here.

ubdbra001: I'm not sure I fully understand why raising the error needs to be outside of the lock to be honest, and so also I'm not really sure if splitting the lock has a negative impact. If I recall correctly, I never managed to hit the race condition that was the impetus for the change, so I'm not sure if I'd be able to test this approach to see if it does/doesn't break something.

> I worry we might break some functionality if we move away from BaseException without extensive testing to see if there's any other errors beyond KeyError that should be caught here.

This is primarily why I have left it be, though it may be worth adding some code to highlight the error if it's not one we're expecting, e.g.:
```python
try:
    # do stuff
except BaseException as e:
    if type(e) is not KeyError:
        print(f"unexpected error: {e}" )

    # Do the rest of the stuff
```

twVolc: > I'm not sure I fully understand why raising the error needs to be outside of the lock to be honest, and so also I'm not really sure if splitting the lock has a negative impact. If I recall correctly, I never managed to hit the race condition that was the impetus for the change, so I'm not sure if I'd be able to test this approach to see if it does/doesn't break something.

Moving at least one aspect of the code outside of the lock means that the `IFitworker` thread has an opportunity to acquire the lock and update its results. If the lock is held continuously within `PyplisWorker.update_doas_calibration()` we won't ever see updates to `IFitWorker.results` whilst inside the `PyplisWorker` method. There's a lot going on in this function and I wrote it along time ago, so this combined with it being difficult to recreate the bug we kept seeing in Indonesia makes it difficult to know for certain what is needed. But I have a feeling dropping the lock at some point could be useful. It was definitely something that was needed in at least part of this method as the initial tests with the lock acquired for the entirety of the method didn't work well if I remember correctly.

> ```python
> try:
>     # do stuff
> except BaseException as e:
>     if type(e) is not KeyError:
>         print(f"unexpected error: {e}" )
> ```

Yes this sounds like a very good solution, so that we can start to pick any other things up - once logging is sorted this would be a useful thing to track.

ubdbra001: I'm probably being very dense here (so massive apologies!) but I'm still not sure why the `IFitworker` thread needs the opportunity to acquire the lock and update its results here?
This bit of code is part of the `else` section of the while loop, so:
- Either checking has timed out, or
- There is a result with a more recent time stamp, so it's likely that point is missing.

So it'll only hold the lock for one last time before either moving onto interpolation or getting the result and proceeding with it.

I am happy to defer to you on this one and change it to the suggestion if you like, but I just want to double check first.

twVolc: No I think you're right. There's a part of me that thinks I put this in how I did for a reason, having tested with the camera, but looking at it again I can't see why this would be. Like you say, the two possible outcomes suggest there's no need for results to be updated again before moving on anyway, so let's stcik with what you've got.

twVolc:
```
try:
    # do stuff
except BaseException as e:
    if type(e) is not KeyError:
        print(f"unexpected error: {e}" )
```
Shall I wait for this update before merging or do you think this should be a different PR?

ubdbra001: > Shall I wait for this update before merging or do you think this should be a different PR?

I'll add it now, otherwise I'll forget!

PR #241, opened 2024-10-24
add function for calculating ICA masses

### LINKED ISSUE #233: Save ICA masses in emission rate file
There's times where flow_glob fails to find a plume speed (gets inf as it things there is no lag between lines). It's then impossible to pull useful data from the sequence (technically you could look at the flow_nadeau emission rate, then divide by it's plume speed to get the ICA mass, but this is convoluted and won't work when the plume speed is calculated to be 0, which it sometimes is).

I think it would therefore be useful for the line's ICA mass to be saved in the emission rate file, so that a user could manually generate a plume speed and then get to an emission rate themselves by scaling the ICA mass.

ubdbra001:
Fixes #233 

Need to:
- [x] Add function
- [x] Add to `calculate_emission_rates`
- [ ] ~Add to `get_cross_corr_emissions_from_buff`~ - Not needed due to how values are recorded and added to the output
- [x] Calculate total ICA mass for lines
- [x] Add value to results
- [ ] ~Add test?~

Adds function to calculate ICA_mass for each PCA/ICA line, plus the total.
Adds the results to the emission rate output files.

Some reorganisation of `calculate_emission_rates()` to simplify code.

Test skipped until after separation of config methods from `PyplisWorker`

### DIFF
--- a/pycam/io_py.py
+++ b/pycam/io_py.py
@@ -328,11 +328,13 @@ def save_so2_img(path, img, filename=None, compression=0, max_val=None):
     cv2.imwrite(full_path, im2save, png_compression)
 
 
-def save_emission_rates_as_txt(path, emission_dict, only_last_value=False):
+def save_emission_rates_as_txt(path, emission_dict, ICA_dict, only_last_value=False):
     """
     Saves emission rates as text files every hour - emission rates are split into hour-long
     :param path:            str     Directory to save to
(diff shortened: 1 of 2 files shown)


ubdbra001:
Okay, I think this one is ready for you to check too.

twVolc:
This all looks good to me, so when you're happy let me know or feel free to merge it yourself.

PR #245, opened 2024-10-28
Fix for sticky image registration

### LINKED ISSUE #134: Image registration a bit buggy (appears to hold onto old registrations)
I've encountered bugs with image registration a number of times when moving between data and generally doing processing. Sometimes it's been hard to keep track of what exactly I've done to recreate the error, so I've just let it slide, but think I need to document this now as I've found at least one way to easily reproduce a bug:

1. Start software, the GUI loads with default villaricca data, and defaut openCV registration
2. Load new image sequence (I load some Reventador data) - as expected, the registration is poor because it still has the Villaricca loaded.
3. There's a few bugs here 

- If I click OpenCV straight away again,  the folllowing error is thrown:

`Exception in Tkinter callback
  File "C:\Users\thom_\anaconda3\envs\pycam\lib\tkinter\__init__.py", line 1892, in __call__
    return self.func(*args)
  File "C:\Users\thom_\Documents\Work\pycam_dev\PyCamPermanent\pycam\gui\figures_cam.py", line 466, in <lambda>
    command=lambda: self.img_reg_select(self.reg_meth))
  File "C:\Users\thom_\Documents\Work\pycam_dev\PyCamPermanent\pycam\gui\figures_cam.py", line 540, in img_reg_select
    self.pyplis_worker.register_image(**kwargs)
  File "C:\Users\thom_\Documents\Work\pycam_dev\PyCamPermanent\pycam\so2_camera_processor.py", line 1147, in register_image
    self.img_B.img_warped = self.img_reg.register_image(self.img_A.img, self.img_B.img, **kwargs)
  File "C:\Users\thom_\Documents\Work\pycam_dev\PyCamPermanent\pycam\so2_camera_processor.py", line 4066, in register_image
    warped_B = self.cv_warp_img(img_B)
  File "C:\Users\thom_\Documents\Work\pycam_dev\PyCamPermanent\pycam\so2_camera_processor.py", line 4022, in cv_warp_img
    img_warped = cv2.warpAffine(img, self.warp_matrix_cv, (sz[1], sz[0]),
cv2.error: OpenCV(4.9.0) D:\a\opencv-python\opencv-python\opencv\modules\imgproc\src\imgwarp.cpp:2757: error: (-215:Assertion failed) (M0.type() == CV_32F || M0.type() == CV_64F) && M0.rows == 2 && M0.cols == 3 in function 'cv::warpAffine'`

- If I first click No Registration, then click back to OpenCV, the registration works as expected, but if I open View > More Windows > Cross-correlation plot and go to the Nadeau image, I can clearly see that this absorption image doesn't perfectly reflect the Analysis Window absorption image. There is an odd feature that is not correctly alligned - it's looks to me as if this feature is still registered as the old Villaricca registration (i.e. much higher shift upwards than the Reventador registration). See below - there is some white speckled feature in the bottom right of the Analysis Window image that is shifted upwards above the volcano flank in the Nadeau plot.
image
This part might be more of a bug in the Nadeau plot updating rather than image registration, but there's definitely a few areas where image registration doesn't quite seem to respond as expected (others that I haven't been able to specifically recreate yet, but for instance something to do with images not correctly jumping back and forth between the different registration types in the Camera window).

I would suggest keeping this issue to document any other bugs we may find relating to this. 

For context on how the warping is implemented: rather than creating a new `PyplisWorker` attribute called `img_B_warped` I decided to just give `PyplisWorker.img_B` a new attribute called `img_warped`. Perhaps things aren't correctly updated at times or there is an artifact of old warping that hasn't bee updated, for instance in other attributes later in the processing pipeline that stem from `img_B_warped` (e.g `PyplisWorker.vign_corr_B_warped`)

ubdbra001:
Fixes #134


ubdbra001:
@twVolc I think this should now do everything we talked about for the GUI part of the image registration. Have a go with it and let me know if you're happy with it.

twVolc:
Just testing things.

Clicking openCV when it is already selected still brings up a dialog box saying that image registration has changed and that optical flow won't have been updated until the image directory is reloaded. The message is usuall probably useful, but if it's possible to not have it pop up when selecting a control button that is already selected then that would probably be good - I guess it's a little misleading at the moment as nothing actually changes?

twVolc:
Everything else on this works really well. It feels so much more intuitive and just work as expected! Can't see any bugs from my tests. 

The optical flow/flow nadeau plots not getting updated properly are a touch annoying but as we've discussed I think this is a much more complex problem to fix because of the different permutations with a camera being connected. My inclination would be to ignore this issue for this PR, as this PR does fix a number of frustrating bugs and is valuable in itself. If we have time we can revisit the other aspect at a later date, but I don't think we should prioritise this at all.

How does this sound?

ubdbra001:
> Clicking openCV when it is already selected still brings up a dialog box saying that image registration has changed and that optical flow won't have been updated until the image directory is reloaded. The message is usuall probably useful, but if it's possible to not have it pop up when selecting a control button that is already selected then that would probably be good - I guess it's a little misleading at the moment as nothing actually changes? This isn't the end of the world if it's a pain to add though. Just something that came to mind.

Good point, the latest commit should address that.

> Everything else on this works really well. It feels so much more intuitive and just work as expected! Can't see any bugs from my tests.

🥳 

> The optical flow/flow nadeau plots not getting updated properly are a touch annoying but as we've discussed I think this is a much more complex problem to fix because of the different permutations with a camera being connected. My inclination would be to ignore this issue for this PR, as this PR does fix a number of frustrating bugs and is valuable in itself. If we have time we can revisit the other aspect at a later date, but I don't think we should prioritise this at all.
>
>How does this sound?

That sounds fine to me if you're happy with leaving it as is. The info box shown after the registration changes does warn the user about this so at least the behaviour isn't hidden. The only thing I'd say is that we should open a new issue to document this problem so that ii doesn't get forgotten about.

twVolc:
Just checked this update over and it works perfectly now. So I'll merge things as they are. I'll create a new issue for missing out the optical flow update when switching registration.

PR #256, opened 2024-11-22
Pyproject dependencies

christopher-wild:
Instead of having the requirements both in the requirements.txt and the pyproject.toml, we can get pyproject.toml to reference the requirements file. 

This should save us from having to keep both lists up to date with any changes.


ubdbra001 (approved):
Didn't know that was possible, a rather neat way to maintain both files without the need to remember to update both!

PR #257, opened 2024-11-25
Ensure FOV search, fixed FOV, and preloaded calibration produce identical output

### LINKED ISSUE #73: Deviation in results when using DOAS calibration vs Preloaded calibration method
Preloaded calibration method produces effectively the same results, but inludes data for a couple of additional timepoints.
For the Lascar data (2022-05-20 16:05:15 - 17:30:30) there are additional points at: 16:05:15, 16:35:05, 16:35:10.
Notably the 16:35:05 and 16:35:10 points would occur around the time where the DOAS-FOV search would start, I haven't tested to see if this changes if the time when this starts changes.

### LINKED ISSUE #94: Some discrepancies in data points when using fixed FOV vs FOV search
There is much more discussion on this in # 43, and this also probably links to #73 (may even be the same issue really. 

@ubdbra001 I would be interested to know if the saved emission rate data also drop these time stamps, or is it only the saved calibration data that are missing?

ubdbra001:
Fixes #73
Fixes #94 

I think these changes address the discrepancies between post-processing runs where there is an FOV search, the FOV is fixed, and a pre-loaded calibration is used. The results include all time points (except the last one), and the results for these tome points are equivalent to 6 decimal places (which I think is mg/s).

The only remaining discrepancy is with the full calibration files, but from what I can tell these are very small differences, and do not affect the final output. I think it would be possible to iron out these, but If you're happy with the emissions outputs being identical then it's probably not worth the effort.

### DIFF
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -3751,7 +3751,7 @@ class PyplisWorker:
         # I will be able to process it too. Otherwise I can only process all but the final image, since I will need to
         # generate the optical flow)
         if self.idx_current < self.img_buff_size:
-            num_buff = self.idx_current - 1
+            num_buff = self.idx_current
         else:
             num_buff = self.img_buff_size
 
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -2526,6 +2526,32 @@ class PyplisWorker:
         self.got_doas_fov = True
         self.doas_last_fov_cal = self.img_A.meta['start_acq']
 
+    def check_doas_result_avail(self, img_time, retry_length = 10):
+        """ Check whether doas results for a particular time point are available
+
+        :param datetime img_time: Datetime for current image being processed
+        :param int retry_length: Time in seconds to retry getting results for, defaults to 10
+        :return boolean: Is there a result for this image datetime
+        """
+
+        timeout = datetime.datetime.now() + datetime.timedelta(seconds = retry_length)
+        while (datetime.datetime.now() < timeout):
+            with self.doas_worker.lock:
+                
+                # Check to see if img_time is in the results index, and return if it is
+                if np.any(img_time == self.doas_worker.results.index):
+                    return True 
+
+                # If there are newer results then the result is likely missing and so we can return
+                # sooner
+                if np.any(img_time < self.doas_worker.results.index):
+                    return False
+
+            # Pause for a bit to allow other threads to execute before trying again
+            time.sleep(0.5)
+        
+        return False
+
     def update_doas_calibration(self, img_tau=None, force_fov_cal=False):
         """
         Updates DOAS results to include more data, or FOV location is also updated if this is requested and in the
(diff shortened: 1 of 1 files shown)


ubdbra001:
I'm going to run checks to make sure RTP is not affected today, and I'm happy to run anything else you would like to check.

ubdbra001:
Okay, looks like RTP results are different. Mainly because during the phase before the FOV search, DOAS results can be missed.
If you want to integrate this into the upcoming release it may be worth setting fixing this aside for later as a different issue, if not I should have time to try and fix it all in this PR.

twVolc:
Just so that I know I'm understanding correctly, are theDOAS results missed because they arrive after the images?

Perhaps set this as a different issue? I wonder if there should be the option to add a time-lag, where image processing waits for a user-defined time to check whether spectra arrive before processing. This might be overkill though, I don't know. 

Does all of this to some extent link back to the `lock` issue we've been think about for a while?

ubdbra001:
I think that is because the points haven't been added to the results by the time the image processing gets to the corresponding image. I'm not sure if that's because they haven't arrived, or because they haven't been processed yet.

To some extent I think it does, but I don't think it's the main issue here. I will have a look to see if the existing code we have written to deal with this can be moved/refactored to address the issue. But unless it's really easy I don't see getting it done by the end of the week, and I don't want to hold things up.

Just to note the results are not wildly different results. For both the ICA masses and the emission rates we are talking about differences of tens of grams or less. I can provide the output files for you to take a look at if you like?

twVolc:
Ok great. Like you say, that sounds so small that it isn't important. Making a note of it and ignoring this for now is probably the best approach.

If this was happening for every image pair - i.e. we were never finding a DOAS point at the time that the pair is processed, then you'd assume we wouldn't be able to generate a calibration line and the results would be completely different. So it seem that the majority of points are being correctly matched. 

I think I need to look back at this part of code to understand how it works in more detail. I'm wondering whether a DOAS data point can still be re-incorporated into the calibration even if it wasn't used directly for its own respective image pair. i.e does the below happen:

1. Image pair at time X arrive
2. No DOAS point is available for this time yet
3. Calibration is formed from previous 30 minutes (or however long is set) of calibration data
4. Image pair at time Y arrive
5. DOAS data point for time X is now available.
6. Code generates the calibration point for image pair at time X and DOAS point at time X
7. Image pair at time Y is calibrated with the included time X data point (and time Y data point if the DOAS point at time Y is already available)

Does the above happen, or if the data point at time X isn't available when processing that image pair is the calibration point for time X just ignored forever. 

Does that query make sense?

In an ideal world the calibration would go back and get calibration points for all available pairs of images and DOAS points, but I'm not sure if that's what happens in reality. I have a feeling that data point will just be ignored for all susbsequent calibrations.

ubdbra001:
Created new issue for these probelms: #259

twVolc:
Ah ok, so we're not intermittently dropping points as they're arriving (in RTP), we're just dropping a single point at very specific times in the processing procedure? That does seem much less important, but an interesting issue. Not something to get caught up on. I think my worry was more related to regular dropping of data in RTP and how the code deals with that. But perhaps I'm overthinking it.

ubdbra001:
Exactly, it is just these very specific points.
The code is pretty robust outside of these points because it will retry the check for a missing point quite a few times, and alos just move on if a newer point appears.

ubdbra001:
Okay, I have a fix for the RTP issues {#259). Do you think adding it here or waiting for this to be merged and creating a new PR is a better idea?
All results are now identical 🥳

PR #264, opened 2024-12-04
Pathlib fix

### LINKED ISSUE #263: os.path.join mixing path separators in spec_worker (Windows)
In `spec_worker.py` there are multiple os.path.join operations such as [line 437](https://github.com/volcanotech-sw/PyCamPermanent/blob/71a3ae6ecca7fa9331ddaf03adfa84bcdcc50f89/pycam/doas/spec_worker.py#L437). These cause paths provided with forward slashes such as:
'C:/Users/myusername/Downloads/E2E-test-data-20241017T082115Z-001/E2E-test-data/data'

to be joined with a windows `\\` which results in paths like:

'C:/Users/myusername/Downloads/E2E-test-data-20241017T082115Z-001/E2E-test-data/data\\Processed_spec_2024-12-03T113611'

The `os` module seems to be able to handle paths with mixed separators like this, but it has caused problems with the development of the headless branch and is bad practice to keep.

I think swapping to use the `pathlib.Path` module would be an easy and good fix

christopher-wild:
closes #263 

Fixes Windows and Mac/Linux path seporators being mixed by swapping `os.path.join` operations with `Pathlib.Path` operations

### DIFF
--- a/pycam/doas/doas_worker.py
+++ b/pycam/doas/doas_worker.py
@@ -10,6 +10,7 @@ import queue
 import numpy as np
 import matplotlib.pyplot as plt
 import scipy.integrate as integrate
+from pathlib import Path
 from itertools import compress
 from scipy import signal
 from scipy.optimize import curve_fit, OptimizeWarning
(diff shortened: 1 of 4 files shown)


fredsonnenwald:
In #269 has identified some places in the `FTPClient` class where a transition to Pathlib would probably be of benefit - is this a transition we should look at making across the wider code base?

PR #268, opened 2024-12-13
Use ICA line 1 for Nadeau line auto-generation when current ICA line deleted

### LINKED ISSUE #266: Plume Velocity settings window doesn't open if ICA line is removed
I'm loading a config file where the ICA line used for automative `flow_nadeau` orientation is set as the 3rd line. If I then remove that line (i.e. delete it from the SO2 image in the analysis window) and try to open the Plume Velocity Settings, the frame won't load and I get the following error:

```
Exception in Tkinter callback
    return self.func(*args)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\pycam\gui\figures_analysis.py", line 5710, in generate_frame
    self.nadeau_flow.generate_frame(self.windows)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\pycam\gui\figures_analysis.py", line 4789, in generate_frame
    self.run_nadeau_line()
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\pycam\gui\figures_analysis.py", line 4965, in run_nadeau_line
    self.update_pcs_line(draw=False)
  File "C:\Users\gg1tcw\Documents\PythonProjects\PyCamPermanent\pycam\gui\figures_analysis.py", line 4992, in update_pcs_line
    self.pcs_line_plot = self.ax_nad.plot([line.x0, line.x1], [line.y0, line.y1], 'c-')
AttributeError: 'NoneType' object has no attribute 'x0'
```

I think if this happens, i.e. the frame is looking for a line that no longer exists, it should just revert back to the first ICA line.

ubdbra001:
Fixes #266

When ICA line used for flow Nadeau line auto generation is removed the GUI will revert back to ICA line 1.
This PR also limits the spinbox options to available lines, and adds the ability to type the ICA line number and click off or press the return key to change the line used.

### DIFF
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -909,12 +909,13 @@ class ImageSO2(LoadSaveProcessingSettings):
         # Once removed, set the line to None
         self.PCS_lines_list[line_num] = None
 
-        if pyplis_worker.auto_nadeau_line and pyplis_worker.auto_nadeau_pcs == line_num:
+        if pyplis_worker.auto_nadeau_pcs == line_num:
             pyplis_worker.config['auto_nadeau_pcs'] = 0
-            messagebox.showwarning(
-                "Reverting ICA line used for Nadeau line autogeneration",
-                "ICA line used for Nadeau line autogeneration removed\n"
-                "Reverting to ICA line 1 for Nadeau line autogeneration")
+            if pyplis_worker.auto_nadeau_line:
+                messagebox.showwarning(
+                    "Reverting ICA line used for Nadeau line autogeneration",
+                    "ICA line used for Nadeau line autogeneration removed\n"
+                    "Reverting to ICA line 1 for Nadeau line autogeneration")
 
         if update_all:
             # Gather variables
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -912,10 +912,8 @@ class ImageSO2(LoadSaveProcessingSettings):
         if pyplis_worker.auto_nadeau_pcs == line_num:
             pyplis_worker.config['auto_nadeau_pcs'] = 0
             if pyplis_worker.auto_nadeau_line:
-                messagebox.showwarning(
-                    "Reverting ICA line used for Nadeau line autogeneration",
-                    "ICA line used for Nadeau line autogeneration removed\n"
-                    "Reverting to ICA line 1 for Nadeau line autogeneration")
+                print("The ICA line selected for Nadeau line autogeneration was removed. "
+                      "Reverting to ICA line 1 for Nadeau line autogeneration")
 
         if update_all:
             # Gather variables
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -909,6 +909,8 @@ class ImageSO2(LoadSaveProcessingSettings):
         # Once removed, set the line to None
         self.PCS_lines_list[line_num] = None
 
+        if pyplis_worker.auto_nadeau_line and pyplis_worker.auto_nadeau_pcs == line_num:
+            pyplis_worker.config['auto_nadeau_pcs'] = 0
         if update_all:
             # Gather variables
             self.gather_vars()
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -912,6 +912,7 @@ class ImageSO2(LoadSaveProcessingSettings):
         if pyplis_worker.auto_nadeau_pcs == line_num and permanent:
             pyplis_worker.config['auto_nadeau_pcs'] = 0
             if pyplis_worker.auto_nadeau_line:
+                self.root.after_idle(self.show_auto_nadeau_warn)
                 print("The ICA line selected for Nadeau line autogeneration was removed. "
                       "Reverting to ICA line 1 for Nadeau line autogeneration")
 
@@ -924,6 +925,13 @@ class ImageSO2(LoadSaveProcessingSettings):
 
             # Redraw canvas
             self.img_canvas.draw()
+
+    @staticmethod
+    def show_auto_nadeau_warn():
+        messagebox.showwarning(
+            "Reverting ICA line used for Nadeau line autogeneration",
+            "ICA line used for Nadeau line autogeneration removed\n"
+            "Reverting to ICA line 1 for Nadeau line autogeneration")
 
     def change_cmap(self, cmap):
         """Change colourmap of image"""
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -911,6 +911,11 @@ class ImageSO2(LoadSaveProcessingSettings):
 
         if pyplis_worker.auto_nadeau_line and pyplis_worker.auto_nadeau_pcs == line_num:
             pyplis_worker.config['auto_nadeau_pcs'] = 0
+            messagebox.showwarning(
+                "Reverting ICA line used for Nadeau line autogeneration",
+                "ICA line used for Nadeau line autogeneration removed\n"
+                "Reverting to ICA line 1 for Nadeau line autogeneration")
+
         if update_all:
             # Gather variables
             self.gather_vars()
--- a/pycam/gui/figures_analysis.py
+++ b/pycam/gui/figures_analysis.py
@@ -767,7 +767,7 @@ class ImageSO2(LoadSaveProcessingSettings):
         ica_num = self.num_ica
         for ica in self.PCS_lines_list[self.num_ica:]:
             if ica is not None:
-                self.del_ica(ica_num)
+                self.del_ica(ica_num, permanent=True)
                 # self.PCS_lines_list[ica_num] = None   # I think this isn't required as it is done in del_ica
             ica_num += 1
 
@@ -876,7 +876,7 @@ class ImageSO2(LoadSaveProcessingSettings):
         else:
             print('Clicked outside axes bounds but inside plot window')
 
-    def del_ica(self, line_num, update_all=True):
+    def del_ica(self, line_num, update_all=True, permanent = False):
         """Searches axis for line object relating to pyplis line object and removes it
 
         Parameters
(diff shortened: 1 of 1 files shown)


twVolc:
I think there's something that's being executed here repeatedly that shouldn't be. Once I follow the steps to hit the error (it now throws the warning box saying that Pycam is reverting to line 0, as expected) I then get the repetition of this line:

```No artists with labels found to put in legend.  Note that artists whose label start with an underscore are ignored when legend() is called with no argument.``` 

in the console. 

Could this be something to do with the new bindings for `FocusOut` and `Return`?

ubdbra001:
I get two repetitions of this message when I remove a line, is this the same for you or do you get more?

In both cases the messages are generated when the `update_plot` method in the `TimeSeriesFigure` class is called, specifically         `self.axes[0].legend(loc='upper left')` presumably becasue there is no legend?
So, on my end it doesn;t look like it is due to the `FocusOut` and `Return` bindings.

twVolc:
Sorry I'll get to this tomorrow!

twVolc:
I'm getting an issue with the line being continually printed to the console. So there's something going on somewhere that is looping over that I haven't seen prior to this issue fix. I'm not sure why though!

twVolc:
I've also just stumbled upon another way of hitting a bug related to #266. If you have Automatic line generation unticked but the line set as line 2. If you then go to the analysis window and remove line 2 the warning box doesn't appear and auto nadeau line doesn't revert to line 1 (presumably because Auto line generation isn't set so it doesn't hit an issue). But then if you go back to Plume Velocity Settings and tick Automatic line generation with the line still set as 2 there is an error thrown (because the line doesn't exist). I guess the reverting back to line 1 needs to happen regardless of if Automatic line generation is selected, to ensure that a user doesn't then accidentally select the non-existent line 2.

ubdbra001:
> I'm getting an issue with the line being continually printed to the console. So there's something going on somewhere that is looping over that I haven't seen prior to this issue fix. I'm not sure why though!

Okay, so not the same as what I'm seeing. You're seeing this immediately after you remove the second ICA line in the analysis window. Are you using the up/down arrows? Is it showing the error message?



> I've also just stumbled upon another way of hitting a bug related to #266. If you have Automatic line generation unticked but the line set as line 2. If you then go to the analysis window and remove line 2 the warning box doesn't appear and auto nadeau line doesn't revert to line 1 (presumably because Auto line generation isn't set so it doesn't hit an issue). But then if you go back to Plume Velocity Settings and tick Automatic line generation with the line still set as 2 there is an error thrown (because the line doesn't exist). I guess the reverting back to line 1 needs to happen regardless of if Automatic line generation is selected, to ensure that a user doesn't then accidentally select the non-existent line 2.

Okay, that should be easy to modify. In the case where the Auto line generation is unticked do you still want to show the warning popup?

twVolc:
> Okay, so not the same as what I'm seeing. You're seeing this immediately after you remove the second ICA line in the analysis window. Are you using the up/down arrows? Is it showing the error message?

Just checked this. It happens after I've clicked ok on the pop-up. So it isn't until after the pop-up is closed that the warning appears and is then continually printed every 0.5 seconds or so. I can try to do some debugging to hunt down where this is happening.

I just click the down arrow to shift the line from 2 to 1.

ubdbra001:
> Just checked this. It happens after I've clicked ok on the pop-up. So it isn't until after the pop-up is closed that the warning appears and is then continually printed every 0.5 seconds or so. I can try to do some debugging to hunt down where this is happening.
> 
> I just click the down arrow to shift the line from 2 to 1.

I'll see if I can reproduce it on my end. I may ask for your help debugging if I can't.

ubdbra001:
> I'll see if I can reproduce it on my end. I may ask for your help debugging if I can't.

No luck on my end, it shows the `No artists with labels found to put in legend.` message, and will show it multiple times if I hold down my mouse button, but it will always stop shortly after I've released the mouse button (same goes for the down arrow key).

There are no additional actions bound to the spinbox, so this isn't due to that.

twVolc:
Interesting that we're seeing different behaviour! I'll try it in anaconda outside of pycharm and see if that changes for me. Otherwise I'll do a bit of a hunt to try to understand why mine is doing this.

twVolc:
image
I get it in Anaconda too...

ubdbra001:
What happens when you have 3 lines and you press to remove the 3rd?

twVolc:
> What happens when you have 3 lines and you press to remove the 3rd? Does it end up removing all except line 1?

Ah yes it does. It shouldn't do that. It should just remove the 3rd line so the second line should remain on the figure. It doesn't matter which the auto line reverts to, that can be 1 or 2 in that case, but there should still be 2 lines in the analysis window either way.

Could this be related to the issue in some way? I'm guessing not, but fix this then checking might be worthwhile

ubdbra001:
I think the button is getting "stuck" and so is just running the `del_ica` method over and over (hence the message reappearing). Would you be able to switch to the dev25.1 branch and see if the behaviour happens there? If it doesn't then it may be that the warning box causes the button to "stick"

twVolc:
> I think the button is getting "stuck" and so is just running the `del_ica` method over and over (hence the message reappearing). Would you be able to switch to the dev25.1 branch and see if the behaviour happens there? If it doesn't then it may be that the warning box causes the button to "stick"

Yep on dev25.1 I can add and remove ICA lines as expected, it doesn't continue deleting additional lines, whether Auto nadeau line is selected or not.

twVolc:
On this branch it does only delete the extra lines when Auto line is selected and the warning box appears. So I think you're right, it's directly related to the warning box in some way. Or the code associated with that part at least.

twVolc:
image
You're right. Commenting this out I then only get the message appearing twice, not continuously. The line number still reverts back to 1, so that confirms it's working and doing what we're expecting.

twVolc:
Going from 3 down to 2 also works as expected - the 2nd ICA line stays in the SO2 image, the Auto nadeau line reverts back to line 1

ubdbra001:
Okay, I'll have a tinker with the warning message to see if I can stop it interfering with the button.

twVolc:
My guess would be this could have something to do with the __draw_canv__() function which refreshes and redraws the SO2 image whenever requested (and at a `refresh_rate` defined at the top of `figures_analysis.py`). Although I don't know why this would keep being invoked in relation to the warning box.  I might be totally wrong and this is completely unrelated...

If this is causing at lot of hassle we can just remove the warning box for now, maybe put something in the logger instead. It's not worth losing a lot of time to such a small function.

ubdbra001:
I've moved the warning after the `update_all` if-statement, which might help. My reasoning is that if it does all the drawing before the warning box is called it shouldn't interfere, but I'm not massively confident about that....
Would you mind having a go to see if that helps?

twVolc:
> I've moved the warning after the `update_all` if-statement, which might help. My reasoning is that if it does all the drawing before the warning box is called it shouldn't interfere, but I'm not massively confident about that.... Would you mind having a go to see if that helps?

Unfortunately that hasn't fixed it. Have you got any idea why this is happening on my computer but not yours?

ubdbra001:
I've updated it so that the warning box has been replaced by a print statement.
I'll spend a little bit more time seeing if I can find a way to address this. My preference is to have an explicit warning, but as you say it's probably not worth spending a tonne of time on.

twVolc:
> > Unfortunately that hasn't fixed it. Have you got any idea why this is happening on my computer but not yours?
> 
> Ah, a pity. Not really, the only obvious thing I can think of is if you're using a touch pad on a laptop?

I'm on a mouse unfortunately.

> We could ask Chris and Fred if they can replicate it, but that may be sinking too much time into this small thing.

I think let's just remove the warning box and finish this issue. Maybe just add the warning message as a print statement/whatever method you're using for logging? That way at least it isn't silently ignored. In general a user should realise that if they remove a line that line can't be used for Auto Nadeau, so whilst the warning box is nice it isn't critical.

twVolc:
> I've updated it so that the warning box has been replaced by a print statement.
> I'll spend a little bit more time seeing if I can find a way to address this. My preference is to have an explicit warning, but as you say it's probably not worth spending a tonne of time on.

Ok great, let me know when you're happy to merge. But yeh I wouldn't spend much more time on this.

ubdbra001:
I think I've spotted a couple of other places where this doesn't work quite as expected. e.g. if you flip the normal for the ICA line selected for Nadeau autogen, then it will reset the line used to 1 because when flipping the line is deleted and recreated.
I assume that this wouldn't be the desired behaviour.

twVolc:
> I think I've spotted a couple of other places where this doesn't work quite as expected. e.g. if you flip the normal for the ICA line selected for Nadeau autogen, then it will reset the line used to 1 because when flipping the line is deleted and recreated. I assume that this wouldn't be the desired behaviour.

Ah good spot, no this wouldn't be desired behaviour. But again, it's not the total end of the worl if this is hard to correct, it's a very  unlikely edge case as really once you create a line and have the normal pointed the right way you're very unlikely to flip it at a later date. Does this issue also happen if you redraw that line then? As presumably the redrawing deletes the old one in the same wya that flipping the normal does. This would make the issue more likely to appear as redrawing a line is much more likely.

ubdbra001:
It does, looking at the code it will currently happen:
- When the line is removed
- When a line is redrawn or overwritten by loading a line from file
- When the normal is flipped

My understanding is that you'd only like the auto gen value to be reset in the first case, is that right?

If we're going to cover this functionality then I think it would be preferable to make sure these edge cases are handled. Though I'm happy to push it down my to-do list until after more important features are completed (e.g. logging), if that is preferable.

twVolc:
Yes my thinking is we only want to drop to a different line, if the current line is actually removed so it's impossible for that line to be used. In all other edit cases the line should stay as the same number.

I think yes push it down the to-do list for now as it's maybe something I can look into too. The logging etc are things that's I'll struggle with more so it's bestto have you working on that.

ubdbra001:
Okay, I've switched this to a draft and will concentrate on the logging.

twVolc:
Sorry I've only just got to this but I've just checked it and it has worked! No repetitive messages in the console and the line reverts to line 1. If 3 or more lines exist only the one line is deleted, rather than iterating through each line to leave only 1 at the end.

twVolc:
This is also working as expected when the ICA normal is flipped and when the line is redrawn (i.e. the auto nadeau line number isn't changed - it's as if the line wasn't deleted during editing) .

ubdbra001:
Okay, if you're happy with it then feel free to merge!

Issue #262, opened 2024-12-02
GUI code attempts to launch sudo


fredsonnenwald:
When launching the GUI under Linux I got a sudo prompt to enter my password to run commands as root that I was not at all expecting. I know the GUI is not listed as supporting Linux, but I believe this could also be happen if the the GUI is launched from the Windows Subsystem for Linux (WSL) on a Windows PC. Depending on the WSL configuration, these sudo commands could be running silently in the background with unintended consequences.

The sudo commands come from the `FTPClient` used by the GUI looking for the SSD mount location (`mount_path`) on the Pi, and doing this by running the `StorageMount` code designed run to manage the SSD on the Pi.

As `mount_path` is not part of any configuration file, and is not set anywhere other than in the intialisation of `StorageMount` I suggest changing `mount_path` into a class property that `FTPClient` can access without initialising `StorageMount` and thereby avoid calling on sudo.

Stack trace:
```
$ python run_pycam.py 
/home/fcs/Desktop/work/volcanoes/PyCamPermanent/pycam/controllers.py:30: UserWarning: Working on machine without seabreeze, functionality of some classes will be lost
  warnings.warn(
/home/fcs/Desktop/work/volcanoes/PyCamPermanent/pycam/controllers.py:38: UserWarning: Working on machine without Picamera2, functionality of some classes will be lost
  warnings.warn(
[sudo] password for fcs: 
Traceback (most recent call last):
  File "/home/fcs/Desktop/work/volcanoes/PyCamPermanent/run_pycam.py", line 1, in <module>
    from pycam.gui.pycam_gui import run_GUI
  File "/home/fcs/Desktop/work/volcanoes/PyCamPermanent/pycam/gui/pycam_gui.py", line 14, in <module>
    from pycam.gui.menu import PyMenu
  File "/home/fcs/Desktop/work/volcanoes/PyCamPermanent/pycam/gui/menu.py", line 5, in <module>
    from pycam.gui.network import ConnectionGUI, instrument_cmd, run_pycam
  File "/home/fcs/Desktop/work/volcanoes/PyCamPermanent/pycam/gui/network.py", line 5, in <module>
    import pycam.gui.cfg as cfg
  File "/home/fcs/Desktop/work/volcanoes/PyCamPermanent/pycam/gui/cfg.py", line 9, in <module>
    from pycam.networking.FTP import FTPClient, CurrentDirectories
[...]

PR #269, opened 2024-12-17
change mount path to StorageMount class variable (#262)

fredsonnenwald:
This PR changes `StorageMount` to have a class variable that contains the location of the mounted SSD on the Raspberry Pi. This lets the `FTPClient` class access this information without creating an instance of `StorageMount` when loading the GUI.

This PR addresses issue #262  where loading the GUI prompted for a password in `sudo`, which occurred as a result of  `StorageMount.__init__()` calling `StorageMount.find_dev()`. `find_dev()` is only applicable to running on the Raspberry Pi and should not be executed by the GUI.

### DIFF
--- a/pycam/networking/FTP.py
+++ b/pycam/networking/FTP.py
@@ -305,7 +305,7 @@ class FTPClient:
     :param network_info:    dict                Contains network parameters defining information for FTP transfer
     """
 
-    def __init__(self, img_dir, spec_dir, network_info=None, storage_mount_data_path=StorageMount.mount_path):
+    def __init__(self, img_dir, spec_dir, network_info=None, storage_mount_data_path=StorageMount.data_path):
         # TODO Need a way of changing the IP address this connects to - this should be linked to sock ip somehow
 
         self.refresh_time = 1   # Length of time directory watcher sleeps before listing server images again
--- a/pycam/networking/FTP.py
+++ b/pycam/networking/FTP.py
@@ -305,13 +305,13 @@ class FTPClient:
     :param network_info:    dict                Contains network parameters defining information for FTP transfer
     """
 
-    def __init__(self, img_dir, spec_dir, network_info=None, storage_mount=StorageMount()):
+    def __init__(self, img_dir, spec_dir, network_info=None, storage_mount_data_path=StorageMount.mount_path):
         # TODO Need a way of changing the IP address this connects to - this should be linked to sock ip somehow
 
         self.refresh_time = 1   # Length of time directory watcher sleeps before listing server images again
         self.cam_specs = CameraSpecs()
         self.spec_specs = SpecSpecs()
-        self.storage_mount = storage_mount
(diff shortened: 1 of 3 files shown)


pycam/networking/FTP.py
```diff
@@ -305,13 +305,13 @@ class FTPClient:
     :param network_info:    dict                Contains network parameters defining information for FTP transfer
     """
 
-    def __init__(self, img_dir, spec_dir, network_info=None, storage_mount=StorageMount()):
+    def __init__(self, img_dir, spec_dir, network_info=None, storage_mount_data_path=StorageMount.mount_path):
```
twVolc, line 308: I think this wants to be `StorageMount.data_path` rather than `.mount_path`, so that it points to the data directory rather than solely the  mount path?

pycam/utils.py
```diff
@@ -302,9 +302,12 @@ class StorageMount:
     """
     Basic class to control the handling of mounting external memory and storing details of mounted drive
     """
-    def __init__(self, mount_path='/mnt/pycam/', dev_path=None):
+    mount_path = '/mnt/pycam/'
+
```
twVolc, line 306: `data_path` might also need breaking out of `__init__()` too if taking this approach?

pycam/networking/FTP.py
```diff
@@ -693,7 +693,7 @@ def full_ssd_download(self, lock='.lock', delete=False):
         num_files = 0
         for dir_num, date_dir in enumerate(dir_list):
             # Set directory
-            current_dir = self.storage_mount.data_path + '/' + date_dir
+            current_dir = self.storage_mount_data_path + '/' + date_dir
```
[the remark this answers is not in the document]
twVolc: I think @christopher-wild has been moving to `PathLib` recently as it deals with different slashes in a better way?

pycam/utils.py
fredsonnenwald: Yes, I hadn't spotted that. Should be corrected in 4e46797390918c914fcbc74ab4108ad063a65926

pycam/networking/FTP.py
fredsonnenwald: Yes, should be corrected in 4e46797390918c914fcbc74ab4108ad063a65926

pycam/networking/FTP.py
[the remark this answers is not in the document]
fredsonnenwald: Yes there's PR #264 moving some things to Pathlib. It's a more robust way to do handle things to do with files.

There are some places in `FTPClient` that are messing around with paths in weird ways like https://github.com/volcanotech-sw/PyCamPermanent/blob/4e46797390918c914fcbc74ab4108ad063a65926/pycam/networking/FTP.py#L481 that would benefit from a switch to Pathlib. However, I think it could better to hold off on making those changes in `FTPClient` until the instrument code is in a better position to test those changes.

PR #271, opened 2024-12-18
Use forward slashes for all file separators

fredsonnenwald:
In several places the code has hard coded backslashes that I can confirm Linux is unhappy with, e.g., `Error: [Errno 2] No such file or directory: '.\\pycam\\conf\\processing_setting_defaults.yml'`

Some additional testing seems to suggest that Python on Linux does not like back slashes at all for file separators and that Pathlib is not a magic fix for that. Some testing on Windows suggests Python on Windows just doesn't care. Forward slash or back slash seem to work equally well, as well as both together, e.g., 

```
>>> t = Path('c:/a folder\\a file.txt')
>>> t.exists()
True
```

To have maximum compatibility I suggest any hard coded paths should use forward slash file separators. End users on Windows should not notice a difference, and if they decided to use a customised config with black slashes that would be compatible. End users on Linux would be able to use the software (provided all else worked OK) without having to manually change black slashes to forward slashes.

This is not a replacement for switch to use Pathlib where appropriate (#264), rather this patch does just enough to get the UI loading on Linux.


christopher-wild (approved):
Interesting to learn that Windows Python is so flexible and that Pathlib can't convert a windows style path to a linux compatible one!

This change should make the paths much simpler and also help with the headless bits 🙌

ubdbra001 (approved):

Issue #254, opened 2024-11-18
pywin32 requirement blocks Linux environment setup - specify for Windows 


fredsonnenwald:
`requirements.txt` asks for `pywin32` [here](https://github.com/twVolc/PyCamPermanent/blob/ccd5085f57ef4a63b52af450f4be25bd62d89a10/requirements.txt#L21), which I think is used in `directory_watcher.py`.

`pywin32` blocks `pip install -r requirements.txt` on Linux as `pywin32` is Windows only.

I think we can change that line in `requirements.txt` to `pywin32; sys.platform == 'win32'` so that it tries to install on Windows only as per https://peps.python.org/pep-0345/#environment-markers. Does this sound sensible?

christopher-wild:
Possibly a dumb question, but will that work on 64bit windows as well do you know?

fredsonnenwald:
Hurrah for finding, so far, 3 relevant PEPs...

- https://peps.python.org/pep-0345/#environment-markers
- https://peps.python.org/pep-0496/
- https://peps.python.org/pep-0508/#environment-markers

It looks like `platform_system=Windows` might be a better option, but otherwise `sys_platform` (note apparently it's now _ not .) could take a value returned by `python3 -c 'import sys; print(sys.platform)'`.

PR #273, opened 2024-12-18
only require pywin32 on windows

fredsonnenwald:
The win32 API is used as part of the directory watching code. This looks to have conditional logic in place for if the code is running on Windows and should use pywin32 or if it's running on Linux and should use pyinotify instead.

https://github.com/volcanotech-sw/PyCamPermanent/blob/71a3ae6ecca7fa9331ddaf03adfa84bcdcc50f89/pycam/directory_watcher.py#L37-L52

This PR makes the pywin32 dependency similarly conditional, so that `pip install -r requirements.txt` can be run on Linux as suggested in #254.

I've tested this on Ubuntu Linux 20.04 and Windows 10 and had the expected behaviour on Linux with pip reporting `Ignoring pywin32: markers 'sys_platform == "win32"' don't match your environment` and pywin32 being installed on Windows and being able to run `import win32` without error.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ srtm.py
 watchdog
 parameterized
 dropbox
-pywin32
+pywin32 ; sys_platform == 'win32'
 pyplis @ git+https://github.com/volcanotech-sw/pyplis.git@sources_fix
 pydoas @ git+https://github.com/volcanotech-sw/pydoas.git
 iFit @ git+https://github.com/volcanotech-sw/iFit.git@make-pip-install
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,6 +18,7 @@ watchdog
 parameterized
 dropbox
 pywin32 ; sys_platform == 'win32'
+pyinotify ; sys_platform == 'linux'
 pyplis @ git+https://github.com/volcanotech-sw/pyplis.git@sources_fix
 pydoas @ git+https://github.com/volcanotech-sw/pydoas.git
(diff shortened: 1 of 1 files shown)


twVolc:
Is there a way to similarly install `pyinotify` if on a linux machine? There's probably loads of other issues with getting the GUI fully working on linux, but just thinking whilst we're solving this issue it might make sense to get the `linux` version of the `DirectoryWatcher` working, or at least all of the correct dependencies installed.

fredsonnenwald:
Yes we can install pyinotify the same way. I've added that in 37ac076a85445709484f42101a22e9133174e41a. When running `pip install -r requirements.txt` on Windows it now reports

```
Ignoring pyinotify: markers 'sys_platform == "linux"' don't match your environment
```

and on Linux it installs and `import pyinotify` runs without error.

fredsonnenwald:
I've checked macOS and there's yet another OS dependent library for file or directory watching that would be needed, [MacFSEvents](https://pypi.org/project/MacFSEvents/). Something for some time in the future maybe.

twVolc:
Excellent, thanks! 

> I've checked macOS and there's yet another OS dependent library for file or directory watching that would be needed, [MacFSEvents](https://pypi.org/project/MacFSEvents/). Something for some time in the future maybe.

Yes let's ignore Mac for now... or maybe forever... I don't think the directory watcher code is currently setup at all for that, so incorporating it would be a much bigger job and probably isn't necessary.

PR #279, opened 2025-01-22
Issue 275 fit ils

twVolc:
@ubdbra001 The only thing I haven't quite got sorted is the warning box doesn't seem to pop up if you load a config file that doesn't have `include_ils_fit` attribute in it. Can you think why this might be?

Other than that this works well - you can toggle the ILS fit on and off in Prcoessing Settings > Calibration > DOAS calibration using the check box next to the ILS figure. DOAS results are slightly different if you process the default sequence with this on and off.

### DIFF
--- a/pycam/conf/processing_setting_defaults.yml
+++ b/pycam/conf/processing_setting_defaults.yml
@@ -56,6 +56,7 @@ auto_bg_cmap: 0
 # DOAS calibration and DOAS-AA calibration
 doas_method: ifit
 ILS_path: .\pycam\tests\test_data\test_spectra\2019-07-03_302nm_ILS.txt
+include_ils_fit: false
 maxrad_doas: 1.0
 remove_doas_mins: 30
 doas_recal: 1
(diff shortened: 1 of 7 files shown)


twVolc:
> @ubdbra001 The only thing I haven't quite got sorted is the warning box doesn't seem to pop up if you load a config file that doesn't have `include_ils_fit` attribute in it. Can you think why this might be?

I'm wondering if I've mis-remembered how this works now, as I've tried removing a few other config parameters from a config file and loading it through File > Load > Load config file and I'm not getting a warning box pop up ever. Are there only some specific attributes which the warning box appears for? I can get it to appear when commenting out some of the directory paths, but not most of the other settings, was this intentional?

ubdbra001:
Currently there are two types of warning for when a config is loaded, both related to paths:
1. If the path isn't in the config then you get a warning and the old path is kept.
2. If the path is in the config, but doesn't actually exist then an error is thrown and the new config isn't loaded.

As far as I recall other parameters in the config are not checked. This is probably something we should address, but it would make sense to do this if/when we get to break out the config handler.

twVolc:
> Currently there are two types of warning for when a config is loaded, both related to paths:
> 
> 1. If the path isn't in the config then you get a warning and the old path is kept.
> 2. If the path is in the config, but doesn't actually exist then an error is thrown and the new config isn't loaded.
> 
> As far as I recall other parameters in the config are not checked. This is probably something we should address, but it would make sense to do this if/when we get to break out the config handler.

Ok this makes sense. In that case I'll merge this as I'm confident it works properly.

PR #283, opened 2025-01-27
Add logging to DOAS related modules 

ubdbra001:
This is the first part of adding logging, and focussed on `SpecWorker` and `IFitWorker`.

As it stands these are logging to the console, and also writing to a general logging file (found in your home directory + `/pycam_logs/root.log`, e.g.: `C:\Users\Daniel\pycam_logs\root.log`), but everything is being written at the `INFO` level (see [here](https://docs.python.org/3.12/library/logging.html#logging-levels) for more details about the levels).

You don't want all of these messages to appear in the console, to do this I can either:
1. Set the messages you want to hide to the `DEBUG` level
2. Set the level filter to `WARNING` and set the messages you want to keep to warning
3. Set up a level between `INFO` and `WARNING`, specifically for processing related messages you want to see in the console.

Option 3 feels like the one that fits best to me (the message you don't want to see aren't really debug info, and the ones you do aren't really warnings), but it will take a little more work to set up. What do you think?

It would also be good if you could identify a couple of messages that you'd like to appear and a couple you'd like to hide from the console, just so I can get an idea.

### DIFF
--- a/pycam/doas/ifit_worker.py
+++ b/pycam/doas/ifit_worker.py
@@ -866,6 +866,10 @@ class IFitWorker(SpecWorker):
 
                 # Stop processing thread when we stop watching the directory
                 self.q_spec.put(self.STOP_FLAG)
+
+                LoggerManager.remove_file_handler(self.SpecLogger, self.log_path)
+                LoggerManager.remove_file_handler(self.SpecDirWatchLogger, self.log_path)
+                LoggerManager.delete_file_handler(self.log_path)
             else:
                 self.SpecDirWatchLogger.warning('No directory watcher to stop')
         else:
--- a/pycam/doas/ifit_worker.py
+++ b/pycam/doas/ifit_worker.py
@@ -472,8 +472,8 @@ class IFitWorker(SpecWorker):
                                                                        wavelengths=self.wavelengths,
                                                                        spectra=self.plume_spec_shift,
                                                                        spec_time=self.spec_time)
-                    self.SpecLogger.info(f'Fit val uncorrected: {fit_0.params['SO2'].fit_val}')
-                    self.SpecLogger.info(f'Fit val corrected: {df_lookup['SO2'][0]}')
+                    self.SpecLogger.info(f'Fit val uncorrected: {fit_0.params["SO2"].fit_val}')
+                    self.SpecLogger.info(f'Fit val corrected: {df_lookup["SO2"][0]}')
                     # Update SO2 value for the best SO2 value we have
                     if not np.isnan(df_lookup['SO2'][0]):
                         fit_0.params['SO2'].fit_val = df_lookup['SO2'][0]
@@ -524,7 +524,8 @@ class IFitWorker(SpecWorker):
 
         # Loop through directory plume images processing them
         for i in range(len(self.spec_dict['plume'])):
-            self.SpecLogger.info(f'Processing spectrum {i+1} of {len(self.spec_dict['plume'])}')
+            self.SpecLogger.info(f'Processing spectrum {i+1} of {len(self.spec_dict["plume"])}')
+            
 
             self.wavelengths, self.plume_spec_raw = load_spectrum(os.path.join(self.spec_dir,
                                                                                self.spec_dict['plume'][i]))
@@ -1341,7 +1342,7 @@ class IFitWorker(SpecWorker):
                 # Change analyser values
                 self.analyser0.params['LDF'].set(value=ldf_best)
                 self.analyser1.params['LDF'].set(value=ldf_best)
-            self.SpecLogger.info(f'LDF: {self.analyser0.params['LDF'].value}')
+            self.SpecLogger.info(f'LDF: {self.analyser0.params["LDF"].value}')
 
             # Fit spectrum for in 2 fit windows
             fit0 = self.analyser0.fit_spectrum([wavelengths[i], spectra[i]], calc_od=['SO2', 'Ring', 'O3'],
--- a/pycam/doas/ifit_worker.py
+++ b/pycam/doas/ifit_worker.py
@@ -711,14 +711,11 @@ class IFitWorker(SpecWorker):
         self.log_path = os.path.join(self.doas_outdir, "SpecWorker.log")
 
         LoggerManager.add_file_handler(self.SpecLogger, self.log_path)
-        LoggerManager.add_file_handler(self.SpecDirWatchLogger, self.log_path)
 
         # Begin processing
         self._process_loop(continuous_save=False)
 
-        LoggerManager.remove_file_handler(self.SpecLogger, self.log_path)
-        LoggerManager.remove_file_handler(self.SpecDirWatchLogger, self.log_path)
-        LoggerManager.delete_file_handler(self.log_path)
+        LoggerManager.remove_file_handler(self.SpecLogger, self.log_path, delete=True)
 
     def _process_loop(self, continuous_save=True):
         """
--- a/pycam/doas/ifit_worker.py
+++ b/pycam/doas/ifit_worker.py
@@ -20,7 +20,6 @@ from scipy.interpolate import griddata
 from matplotlib.pyplot import GridSpec
 from shapely.geometry import Point, Polygon     # For Varnam light dilution
 from shapely.strtree  import STRtree
-from pycam.directory_watcher import create_dir_watcher
 from pycam.setupclasses import SpecSpecs, FileLocator
 from pycam.io_py import load_spectrum
 from pycam.ifit_ld import lookup
@@ -854,27 +853,6 @@ class IFitWorker(SpecWorker):
     def stop_sequence_processing(self):
         """Stops processing a sequence"""
         self.q_stop.put(self.STOP_FLAG)
-
-    def start_watching(self, directory, recursive=True):
-        """
-        Setup directory watcher for images - note this is not for watching spectra - use DOASWorker for that
-        Also starts a processing thread, so that the images which arrive can be processed
-        """
-        if self.watching:
-            self.SpecDirWatchLogger.warning(
-                f'Already watching for spectra: {self.transfer_dir}\n'
-                f'Please stop watcher before attempting to start new watch. This isssue may be '
-                f'caused by having manual acquisitions running alongside continuous watching'
-            )
-            return
-        self.watcher = create_dir_watcher(directory, recursive, self.directory_watch_handler)
-        self.watcher.start()
-        self.transfer_dir = directory
-        self.watching = True
-        self.SpecDirWatchLogger.info(f'Watching {self.transfer_dir[-30:]} for new spectra')
-
-        # Start the processing thread
-        self.start_processing_thread()
 
     def stop_watching(self):
         """Stop directory watcher and end processing thread"""
--- a/pycam/doas/ifit_worker.py
+++ b/pycam/doas/ifit_worker.py
@@ -14,6 +14,7 @@ import datetime
 import numpy as np
 import pandas as pd
 import matplotlib.pyplot as plt
+from pathlib import Path
 from itertools import compress
 from tkinter import filedialog
 from scipy.interpolate import griddata
@@ -766,12 +767,22 @@ class IFitWorker(SpecWorker):
                 spec_time = self.get_spec_time(filename)
 
                 if not self.first_spec and spec_time.day != self.spec_time.day:
-                    self.SpecLogger.info("new day found")
+                    self.SpecLogger.info(f"Day {self.spec_time.strftime('%Y-%m-%d')} finished")
+                    mem_handler.flush()
+
                     self.reset_self()
-                    # Set output dir
-                    self.set_output_dir(working_dir)
 
                 if self.first_spec:
+                    if self.watching:
+                        self.SpecLogger.info(f"Day {spec_time.strftime('%Y-%m-%d')} started")
+                    
+                        # Set output dir
+                        self.set_output_dir(working_dir)
+                        self.log_path = Path(self.doas_outdir).joinpath("SpecWorker.log").as_posix()
+
+                        file_handler = LoggerManager.create_file_handler(self.log_path)
+                        mem_handler = LoggerManager.set_mem_handler_target("SpecWorker", file_handler)
+                    
                     # Save processing params
                     self.save_doas_params()
                     header = True
@@ -846,6 +857,9 @@ class IFitWorker(SpecWorker):
                 self.fig_spec.update_clear()
             else:
                 self.SpecLogger.debug(f'Spectrum type not recognised: {pathname}')
+            
+            if self.watching:
+                mem_handler.flush()
 
     def stop_sequence_processing(self):
         """Stops processing a sequence"""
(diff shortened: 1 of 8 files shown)


twVolc:
This is looking good to me. I'm guessing the plan is to incorporate this as a separate PR to `PyplisWorker` logging? Although from #265 it looks like you have it all implemented already?

Is the log saved somewhere? I haven't managed to find that yet. 

It would also be nice to be able to surpress warnings like the `FutureWarning` and `UserWarning` that pop up on start-up and at the start of DOAS processing, if this is possible. It will just tidy the console a little. I don't think the time taken for each processing step in DOAS is necessarily important for the user, so this can be removed - but that being said, I can see it being useful for benchmarking and testing different setups and any updates, so perhaps it just wants commenting out, or if there's a way of logging things at different levels maybe this could be set to a less important level, so it can be surpressed but easily brought back up when developing code - is that something that's possible?

ubdbra001:
Yes, I want to do it piecewise for the two different components. The foundation is all implemented, but the specific will need some work.

> Is the log saved somewhere? I haven't managed to find that yet.

Not currently in this PR, I want to get the console logging right then I'll update it with file based logging. 

> It would also be nice to be able to surpress warnings like the `FutureWarning` and `UserWarning` that pop up on start-up and at the start of DOAS processing,

I want to see if the changes Chris has made to update the libraries for Python 3.12 will deal with those warnings.

> I don't think the time taken for each processing step in DOAS is necessarily important for the user, so this can be removed - but that being said, I can see it being useful for benchmarking and testing different setups and any updates, so perhaps it just wants commenting out, or if there's a way of logging things at different levels maybe this could be set to a less important level, so it can be surpressed but easily brought back up when developing code - is that something that's possible?

Yep, the minimum log level can be specified so that only messages of that level and above will appear in a particular log. E.g. we can set the times as debug messages (the lowest level), and so if you set the console or file logger to watch for the debug level then they'd show, otherwise they'd be hidden.
 I'm not sure how I'd set this up so that it can be changed by the user, but if you think it would be useful then I can have a think.

twVolc:
Great! I wouldn't worry about making log levels for different parts user adjustable, if that's what you mean, but as long as it's possible for us as developers to set a log level when we're running things that would be good. Actually looking at it, maybe the processed statement, that includes the time taken, is the one that the user should see, and the "Got new file" message should be a lower level that isn't visible. We basically don't want to overload the console with too much info for the user.

twVolc:
> ~I had a look at the `futureWarning` and they are not fixed in the 3.12 update.~ It is no longer a problem once I actually rebuilt the environment with 3.12 🤦‍♂️

Haha that's good to know! I'm guessing Chris fixed these when updating the code?

I'll try to get to checking over this ASAP. Is there a way of defining what logging level we want to run the software at, i.e. a location in a file that I can change the level so that different warning either appear or disappear? Just checking so that I can play around with that too if it exists.

ubdbra001:
> Haha that's good to know! I'm guessing Chris fixed these when updating the code?

Either that or the change it was warning about has happened but it doesn't have an impact on PyCam.

> Is there a way of defining what logging level we want to run the software at, i.e. a location in a file that I can change the level so that different warning either appear or disappear? Just checking so that I can play around with that too if it exists.

By default it will show INFO and above in the console, and everything (DEBUG and above) in the file. Currently the easiest way to change this is to adjust the following lines in spec_worker.py:

```python
SpecLogger = LoggerManager.add_logger("SpecWorker", "blue")
SpecDirWatchLogger = LoggerManager.add_logger("SpecDirWatcher", "purple")
```
The `add_logger` method has an optional 'level' parameter, so you can set it lower, e.g.
```python
SpecLogger = LoggerManager.add_logger("SpecWorker", "blue", level = 10)
```
(You can import the `logging` module and use (e.g.) `logging.DEBUG` but setting the [corresponding integer value](https://docs.python.org/3.12/library/logging.html#logging-levels) works file too)

The method for adding the file handler (`add_file_handler`) can be changed in the same way, it appears in both `spec_worker.py` and `ifit_worker.py`

ubdbra001:
I believe so, but it will only be for the `doas_worker` stuff currently. If you do hit something where it doesn't do that let me know.

twVolc:
> Random observation: I've noticed from the logs that the `doas_worker` gets instantiated 3 times on loading. Once properly and again in the `__init__` for both `RefPlot` and `ILSFrame`. The default parameter for the `doas_work` argument in these classes is `DOASWorker()` which creates a new instance of the object when no parameter is supplied.
> 
> Do you think it's worth changing this to use `doas_worker` (the already instantiated class that is imported)? And if so should we do it here in this PR?

It sounds like this would be good to fix - can't see any good reason to instantiate a new object for these aspects. Happy to include the fix in this PR if you can.

All of the logging looks good to me - it seems to highlight a number of places where things are possibly being duplicated (run multiple times where probably necessary) so this will be great for future optimisation (not for now!).

I think you can remove the `Got new file:` line from logging to the console - perhaps set it to debug level as it might be useful at times but it isn't valuable to the typical user as the `Time taken to process...` line gives the details of the spectra being processed along with the time. I think this is a more valuable thing to print.

ubdbra001:
> It sounds like this would be good to fix - can't see any good reason to instantiate a new object for these aspects. Happy to include the fix in this PR if you can.

Done

> I think you can remove the `Got new file:` line from logging to the console

Also Done

twVolc:
Great, thanks! 

Is it possible to have the main log located in `pycam\logs` or is it hard to work with these relative paths? I think it just might make it easier for the user to find if it's held in the installation directory, or is this bad practice?

twVolc:
> The `add_logger` method has an optional 'level' parameter, so you can set it lower, e.g.
> 
> ```python
> SpecLogger = LoggerManager.add_logger("SpecWorker", "blue", level = 10)
> ```

I've tested this and it works well - could be super useful!

ubdbra001:
> Is it possible to have the main log located in `pycam\logs` or is it hard to work with these relative paths? I think it just might make it easier for the user to find if it's held in the installation directory, or is this bad practice?

It is certainly possible to put it there. I was thinking ahead when I chose this location: If you want the package to be pip-installable in the future then the installation directory is going to be tricky to find.

twVolc:
> > Is it possible to have the main log located in `pycam\logs` or is it hard to work with these relative paths? I think it just might make it easier for the user to find if it's held in the installation directory, or is this bad practice?
> 
> It is certainly possible to put it there. I was thinking ahead when I chose this location: If you want the package to be pip-installable in the future then the installation directory is going to be tricky to find.

Ok good point. I thought there was probably a reason you put it there, so that seems well justified - let's stick with how it is.

ubdbra001:
I think this is pretty much ready to go? I've added Chris and Fred as reviewers just in case they have suggestions for changes

pycam/logging/logging_tools.py
```diff
@@ -0,0 +1,118 @@
+import logging
+import colorlog
+from pathlib import Path
+from logging.handlers import TimedRotatingFileHandler
+
+class LoggerManager:
+    """ Class for manging creation of loggers and creation and deltion of handlers
```
christopher-wild, line 7: minor typos with 'manging' and 'deltion'

pycam/logging/logging_tools.py
```diff
@@ -0,0 +1,118 @@
...
+
+        :param str name: Name of the logger to create
+        :param str colour: Colour of the text used for logging messages, defaults to white
+        :param int level: Logging level to set for the StreamHandler, defaults to logging.DEBUG
+        :return logger: Reference to logger
+        """
+
+        if name not in LoggerManager._loggers:
+            logger = logging.getLogger(name)
+            logger.setLevel(logging.DEBUG)
```
christopher-wild, line 27: Quick question just to check this works how I think it does

We're allowing each handler level to be configurable rather than the logger level? And if a handler level is for example `INFO` while the logger level is `DEBUG`, we will still only see `INFO` or higher messages come through the handler?

pycam/doas/spec_worker.py
```diff
@@ -521,8 +525,42 @@ def save_results(self, pathname=None, start_time=None, end_time=None, save_last=
...
+        self.watcher = create_dir_watcher(directory, recursive, self.directory_watch_handler)
+        self.watcher.start()
+        self.transfer_dir = directory
+        self.watching = True
+        self.SpecDirWatchLogger.info(f'Watching {self.transfer_dir[-30:]} for new spectra')
+
+        # Start the processing thread
+        self.start_processing_thread()
+
+    def reset_self(self):
```
christopher-wild, line 562: Should this reset_self be set to an ABC method or removed?

pycam/so2_camera_processor.py
```diff
@@ -56,6 +57,11 @@
     "species_paths"
 ]
 
+# Remove StreamHandlers from pyplis loggers so messages from pyplis just get passed to the
```
christopher-wild, line 60: Whats the motivation for having pyplis log to the root logger?

Also I cant seem to find where pyplis gains a `.logger` and `.print_log`

christopher-wild (commented):
couple of small questions but looks great

pycam/logging/logging_tools.py
ubdbra001, line 7:
```suggestion
    """ Class for managing creation of loggers and creation and deletion of handlers 
```

pycam/doas/ifit_worker.py
```diff
@@ -783,7 +798,8 @@ def _process_loop(self, continuous_save=True):
 
                 time_1 = time.time()
                 self.process_doas()
-                print('IFit Worker: Time taken to process {}: {}'.format(pathname, time.time() - time_1))
+                process_duration = time.time() - time_1
+                self.SpecLogger.info(f'Time taken to process {pathname}: {process_duration:.5f}')
```
fredsonnenwald, line 802: You can put simple arithmetic inside of f-strings, if process_duration isn't used elsewhere suggest that to avoid having another local variable floating around.
```suggestion
                self.SpecLogger.info(f'Time taken to process {pathname}: {time.time() - time_1:.5f}')
```

pycam/so2_camera_processor.py
ubdbra001: > Whats the motivation for having pyplis log to the root logger?

Essentially to hide the messages that get produced by pyplis from the console, but still collect them in the file log.

> Also I cant seem to find where pyplis gains a `.logger` and `.print_log`

They get created [here](https://github.com/volcanotech-sw/pyplis/blob/dd8e9dcd10a0435cbd000019471e9529b147263a/pyplis/__init__.py#L25C19-L25C35) in pyplis and just left on.

Actually, Fred did suggest that it may be worth setting the levels for them to `WARNING` and above  in the console so that any issues can be spotted during running. Do you think that sounds like something you'd be happy with @twVolc ?

pycam/doas/ifit_worker.py
```diff
@@ -822,55 +838,42 @@ def _process_loop(self, continuous_save=True):
...
                 self.fig_spec.update_clear()
             else:
-                print('IFitWorker: spectrum type not recognised: {}'.format(pathname))
+                self.SpecLogger.debug(f'Spectrum type not recognised: {pathname}')
 
     def stop_sequence_processing(self):
         """Stops processing a sequence"""
         self.q_stop.put(self.STOP_FLAG)
 
-    def start_watching(self, directory, recursive=True):
```
fredsonnenwald, line 841: I'm not really following the shuffling of the directory watchers, but I'll trust it at all works.

pycam/so2_camera_processor.py
twVolc: > Actually, Fred did suggest that it may be worth setting the levels for them to `WARNING` and above in the console so that any issues can be spotted during running. Do you think that sounds like something you'd be happy with @twVolc ?

I'm not sure I'm totally clear on what this means? Is this relating to warnings that pyplis generates? I think they can be ignored for the most part.

pycam/doas/spec_worker.py
```diff
@@ -439,8 +445,6 @@ def save_clear_raw(self, filename):
 
     def start_processing_thread(self):
         """Public access thread starter for _processing"""
-        # Reset self
-        self.reset_self()
```
fredsonnenwald, line 443: This doesn't look logging-related

pycam/doas/spec_worker.py
ubdbra001: Setting it up as an ABC method means that any child class needs to implement it, right?

pycam/logging/logging_tools.py
```diff
@@ -0,0 +1,118 @@
...
+    def remove_stream_handlers(logger):
+        """ Removes all stream handlers from logger provided.
+
+        :param logger logger: logger object to remove stream handlers from
+        """
+        for handler in logger.handlers:
+            if isinstance(handler, logging.StreamHandler):
+                logger.removeHandler(handler)
+
+# Set up the root logger
```
fredsonnenwald, line 104: When does this run? Looks like it could get called any time `import logging_tools` is run? Could multiple calls cause conflicts?

fredsonnenwald (commented):
Do you have any guidelines for when to use `Logger.info()` vs `Logger.debug()` etc

pycam/so2_camera_processor.py
ubdbra001: Yep: rather than just putting them all into a file, instead having any message from pyplis above a specific severity show up in the console so it's obvious when warnings, errors, etc are happening. (And they'd also appear in the logs)

pycam/logging/logging_tools.py
ubdbra001: It does run every time the module is imported, but `getLogger()` will create a logger with a supplied name first time around and then just get the logger with the same name on each subsequent call. If the name is left blank then it will get the root logger.

The conditional section below adds a file handler to the root logger but only when there isn't one already, so only on first import.

pycam/so2_camera_processor.py
twVolc: Yep sounds good - as long as those errors/warnings are put into the log file too so that they can be traced later too then this is good with me.

pycam/doas/ifit_worker.py
```diff
@@ -754,12 +769,12 @@ def _process_loop(self, continuous_save=True):
                 spec_time = self.get_spec_time(filename)
 
                 if not self.first_spec and spec_time.day != self.spec_time.day:
-                    print("new day found")
+                    self.SpecLogger.info("new day found")
                     self.reset_self()
-
-                if self.first_spec:
                     # Set output dir
                     self.set_output_dir(working_dir)
```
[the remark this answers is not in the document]
ubdbra001: This is to move the `set_output_dir` method so that it is called when a new days worth of data is found (so a new output dir needs to be created), but not on the initial run, because it will have already been created.

Though looking at it I think there needs to be some code for creating and writing to a new logfile in the new output dir...

ubdbra001:
> Do you have any guidelines for when to use `Logger.info()` vs `Logger.debug()` etc

Nothing concrete, Most of it has been inferred from what Tom and Tom have said about what they'd like, which I think can be summarised as: only essential info associated with a processing run should be in the console. Everything else should be in the log file (and therefore `DEBUG`)

pycam/so2_camera_processor.py
ubdbra001: Okay, I'll tweak this so that the handlers are replaced (so messages have a consistent format) and have a logging level of `WARNING`.

twVolc:
> Nothing concrete, Most of it has been inferred from what Tom and Tom have said about what they'd like, which I think can be summarised as: only essential info associated with a processing run should be in the console. Everything else should be in the log file (and therefore `DEBUG`)

I think for this stage just having the logger set up is the key thing. If we find certain messages would be preferable at different levels the more we use it then that will presumably be very easy to change, so it's not something to look deeply at right now.

pycam/doas/spec_worker.py
christopher-wild: It does, but I'm not sure what other use a function that only contains pass would have

pycam/so2_camera_processor.py
ubdbra001: I went with the default being `ERROR` in the end, `pyplis` produces too many warning messages.
As with all the logging stuff this can be modified by setting the `level` parameter.

ubdbra001:
Thanks for the reviews! I'm happy with this now.
May be worth one last look before it is merged.

christopher-wild (approved):

PR #287, opened 2025-02-19
Headless pyplis

christopher-wild:


pycam/so2_camera_processor.py
```diff
@@ -4052,7 +4086,12 @@ def _processing(self):
         Main processing function for continuous processing
         """
         # Reset self
-        self.reset_self()
+        if self.headless:
+            plot = False
+        else:
+            plot = True
+
+        self.reset_self(reset_plot=plot)
```
ubdbra001, line 4094: I'd suggest reverting this and adding a check for headless at the end of `reset_self`, i.e.:
```python
if reset_plot and not self.headless:
    # Reset time series figure
    self.fig_series.update_plot()
```

`reset_self` gets called when RTP moves from one day to the next, and so currently crashes because it tries to access `fig_series`

christopher-wild:
Dan and I have been working on and testing this branch for the last couple of days. The results for DOAS and Pyplis are now consistent between:

- GUI batch processing
- GUI real time processing
- Headless batch processing
- Headless real time processing

So with some more minor bug fixes I think this branch is ready for review

ubdbra001:
Just tried headless running with FOV search, and for some reason it loses the first 10 mins of data (this is the min search window length so I suspect it's not a coincidence)

PR #292, opened 2025-03-17
Add logging for pyplis module

ubdbra001:
Second part of adding logging to PyCam, this time focussing on `PyplisWorker`. Initially this is just for the console logs, I'll add file logging later.

As before it would be good to get feedback on the levels I've selected for these messages. If you do spot anything else that looks odd then do give me a shout!

### DIFF
--- a/pycam/so2_camera_processor.py
+++ b/pycam/so2_camera_processor.py
@@ -351,7 +351,7 @@ class PyplisWorker:
                 Error: {e}\n
                 Reverting to config file supplied with PyCam""")
 
-            print(self.load_default_conf_errors)
+            self.PyplisLogger.warning(self.load_default_conf_errors)
             
             # If any files not found then retry with the supplied config
             self.load_config(FileLocator.PROCESS_DEFAULTS, "default")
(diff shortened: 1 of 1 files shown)


ubdbra001:
File logging has now been added. If you're happy with the levels for this then I think we're all good.

twVolc:
Is there any way to supress some of the pyplis outputs? For instance on start-up there's a huge number of these lines:
```
No emission rate analysis data available for 0
No emission rate analysis data available for 0
No emission rate analysis data available for 0
```
I haven't investigated where exactly they come from, and the repetition may mean that there's inefficient repetitions in the code, so it might be nice to surpress these lines for users but have them still showing up if running at a debug level, so that devs still see that this is happening. Does that make sense? And is this possible?

ubdbra001:
Yes, this is being generated by the GUI, and bits I'm working on will be for the GUI and other more general PyCam modules

twVolc:
Ah ok great. In that case this looks good. I think the `Updating plot A/B` can probably be set to not show by default, or deleted completely, as this is just a debugging thing that doesn't really have use for the most part. Other than that the logging looks good to me

ubdbra001:
Okay, just dropped that to be a debug message. If you're happy with this then feel free to merge.

PR #293, opened 2025-03-21
Last hardware fixes

fredsonnenwald:
These are the last few little fixes for issues found during final hardware testing before handover:

Issues resolved:

- Better support for higher frame rates


christopher-wild (approved):
Not looked at the .sh as I'm not familiar with it but the Python looks good to me

